Issue #164, opened 2019-06-07
[Discussion/Opt] getBsweep multiprocessing routine


OrtnerMichael:
Currently multiprocessing works only if you have MANY POSITIONS and FEW SOURCES.

We should redo the getBsweep so that it pools ALL THE getB calls!!!

lucasgcb:
I feel we could still optimize this by a good margin, but will take discussion.

PR #178, opened 2019-06-19
Enhance collection getBsweep

lucasgcb:
## Related Issues

#164 

## Notes

This adds handling for when `Collection.getBsweep(multiprocessing=True)` happens to have several objects but too few positions to parallelize efficiently, which ends up costing performance.

Benchmark Code:
https://gist.github.com/lucasgcb/40f5979d9436a671ed2600e429f52fa2

Benchmark was run with getBsweep of 1 Object for 10000 positions and getBsweep of compounding 10000 objects in 1 position.

<details>
<summary> Click Here for Benchmark Results </summary>

As we expected, running getBsweep when there are too few positions resulted in ridiculous processing times. I had to cancel out at 500 Objects since time was getting worse by each step.

image

After the tweak, getBsweep achieved a more acceptable processing time for 10000 objects.
Note how doing this is a more complex operation than performing getB, so there's some room for improvement. 

image

</details>

### DIFF
--- a/magpylib/_lib/classes/collection.py
+++ b/magpylib/_lib/classes/collection.py
@@ -32,6 +32,8 @@ listOfPos=[[x,y,z]] # List of Positions
 #######################################
 # %% IMPORTS
 from copy import deepcopy
+from itertools import repeat
+from multiprocessing import cpu_count
 import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 from mpl_toolkits.mplot3d.art3d import Poly3DCollection
@@ -42,8 +44,10 @@ from magpylib._lib.classes.moments import Dipole
 from magpylib._lib.classes.fieldsampler import FieldSampler
 from magpylib._lib.utility import drawCurrentArrows, drawMagAxis, drawDipole, isDisplayMarker
 from magpylib._lib.utility import addListToCollection, isSource,  addUniqueSource, isPosVector
+from magpylib._lib.utility import initializeMulticorePool
 from magpylib._lib.mathLibPrivate import angleAxisRotation
 from magpylib._lib.mathLibPublic import rotatePosition
+
 
 
 class Collection(FieldSampler):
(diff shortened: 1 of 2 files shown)


OrtnerMichael:
its called 'do not merge', yet there is a pull request ?

lucasgcb:
> its called 'do not merge', yet there is a pull request ?

It's Work In Progress actually, this PR is here in order to run additional automated checks as more commits pertaining this branch come in (if they do come in) :octocat:

lucasgcb:
@OrtnerMichael  I've performed the benchmark on a simulated dual core environment, for the issue we were having the problem seems fixed to me.

Noticed however there is still a higher complexity situation in compounding the analysis several thousand objects, in the results you'll see this clearly as there is disparity between several positions vs 1 object and several objects vs 1 position. 

Benchmark Code:
https://gist.github.com/lucasgcb/40f5979d9436a671ed2600e429f52fa2

Benchmark was run with getBsweep of 1 Object for 10000 positions and getBsweep of compounding 10000 objects in 1 position.

<details>
<summary> Click Here for Benchmark Results </summary>

As we expected, running getBsweep when there are too few positions resulted in ridiculous processing times. I had to cancel out at 500 Objects since time was getting worse by each step.

image

After the tweak, getBsweep achieved a more acceptable processing time for 10000 objects.
Note how doing this is a more complex operation than performing getB, so there's some room for improvement. 

image

OrtnerMichael:
If I understand correctly you now separate two cases:
1. Lots of positions & few sources
2. Few positions & Lots of sources

While this is fine in principle, would it not make more sense for the sake of generality to pool ALL cases in the same way. To collect all instances when the field is calculated in a getBsweep into one big 1D list, then do the calculation by parallelization if required, finally reshape the list to its designated form and return?

lucasgcb:
@OrtnerMichael 
I understand where you're coming from and I thoroughly agree with it, but here are my initial thoughts on why it I didn't choose doing that from the start:

The way multiprocessing works is:

1. Take a function address and a list of parameters, we "pickle" this function
2. "Fork" a new process for each of these parameters, and apply them to the "pickled" function

As far as I know, "priming" several functions to be executed into a list may be just as complicated if not more so than what it is now. This would require listing all the method addresses beforehand and then feeding parameters into each and every one of them.

The issues with this situation is that:
 - We have 2 lists of method/parameters to worry about (field method and positions)
 - Each source object is effectively an address that contains the method address we want
 - We have to have to pass the entire parameter list with the pickled function for it to parallelize
 - We also have a limit to how much we can parallelize (number of cores)

The middle ground I went for was creating two routines that mix parallelization and sequential logic. 

1. If there are more positions than cores, it will take each object sequentially and parallelize the field sample calculation.

2. If there are less positions than cores, it will instead opt to start a process for each object and calculate the field for each position sample sequentially within each core.

---

It would still be preferable in my view to have a constant complexity for both cases as you have mentioned so I'll definitely take a better look to see if I can generalize it in some way. 

However, since this fixes a blatant issue with Collection, I'll see to merging this hotfix and work on refactoring in the meantime. Thanks!

PR #179, opened 2019-06-20
display system fix (ax reference instead of plt)

OrtnerMichael:
@lucasgcb 

please check out these changes.

I notices that there were some functions in utility that are used in displaySystem to draw quivers. However, they were given 'pyplot' as argument rather than the respective axis. In this case matplotlib always gives plt.gca() (which is the current active axis).

With the new option subplotAx in displaySystem it is now possible to activate other axes BEFORE the plotting takes place. so the quiver objects will jump to the last used axis.

This problem should be fixed with this commit, but please double check it!!!

### DIFF
--- a/magpylib/_lib/classes/collection.py
+++ b/magpylib/_lib/classes/collection.py
@@ -620,17 +620,17 @@ class Collection(FieldSampler):
         for d in dipolesList:
             drawDipole(d.position, d.moment,
                        d.angle, d.axis,
-                       SYSSIZE, plt)
+                       SYSSIZE, ax)
 
         if direc is True:  # Draw the Magnetization axes and current directions
-            drawCurrentArrows(currentsList, SYSSIZE, plt)
-            drawMagAxis(magnetsList, SYSSIZE, plt)
+            drawCurrentArrows(currentsList, SYSSIZE, ax)
+            drawMagAxis(magnetsList, SYSSIZE, ax)
 
-        for tick in ax.xaxis.get_ticklabels()+ax.yaxis.get_ticklabels()+ax.zaxis.get_ticklabels():
-            tick.set_fontsize(12)
-        ax.set_xlabel('x[mm]', fontsize=12)
-        ax.set_ylabel('y[mm]', fontsize=12)
-        ax.set_zlabel('z[mm]', fontsize=12)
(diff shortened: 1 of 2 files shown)


OrtnerMichael:
the code that broke the current version was

`
from magpylib import source, Collection
from numpy import array, linspace, meshgrid
from numpy.linalg import norm
from matplotlib import pyplot as plt

plt.rcParams.update({'font.size': 6})

fig1 = plt.figure(figsize=(8, 5))
axsA = [fig1.add_subplot(2,3,i, projection='3d') for i in range(1,4)]
axsB = [fig1.add_subplot(2,3,i) for i in range(4,7)]

fig2 = plt.figure(figsize=(8, 5))
axsA += [fig2.add_subplot(2,3,i, projection='3d') for i in range(1,4)]
axsB += [fig2.add_subplot(2,3,i) for i in range(4,7)]

#position grid
ts = linspace(-6,6,50)
posis = array([(x,0,z) for z in ts for x in ts])
X,Y = meshgrid(ts,ts)

# source definitions
s1 = source.magnet.Box(mag=[500,0,500], dim=[4,4,4])
s2 = source.magnet.Cylinder(mag=[0,0,500], dim=[3,5])
s3 = source.magnet.Sphere(mag=[-200,0,500], dim=5)
s4 = source.current.Line(curr=10, vertices=[(0,-5,0),(0,5,0)])
s5 = source.current.Circular(curr=10, dim=5)
s6 = source.moment.Dipole(moment=[0,0,100])

for i,s in enumerate([s1,s2,s3,s4,s5,s6]):
    
    #plot geometry
    c = Collection(s)
    c.displaySystem(subplotAx=axsA[i],markers=[(6,0,6)])
    axsA[i].plot([-6,6,6,-6,-6],[0,0,0,0,0],[-6,-6,6,6,-6])

    #plot field
    #B = array([s.getB(p) for p in posis]).reshape(50,50,3)
    #axsB[i].pcolor(X,Y,norm(B,axis=2),cmap=plt.cm.get_cmap('coolwarm'))
    #axsB[i].streamplot(X, Y, B[:,:,0], B[:,:,2], color='k', density=1, linewidth=1)
`

OrtnerMichael:
And another small fix in there:
There was a fixed fontsize in display system. This should not be the case. The user should be able to adjust the font size through rc parameters or directly.

lucasgcb:
Looks good to me, documentation is also passing locally.

image

Issue #188, opened 2019-07-02
[REQUEST] magnet class improvement


OrtnerMichael:
TRANSCRIPT FROM ALEX:

here is a brief example

example class:

![image](https://user-images.githubusercontent.com/47384012/60508211-4b2bcf80-9cca-11e9-800b-76c311fbed23.png)

You define a new magnet like this:
 
![image](https://user-images.githubusercontent.com/47384012/60508244-60086300-9cca-11e9-9eae-9c1a06f7de2e.png)

with the __repr__ method there is a useful representation of the object (rather than direct reference to the object:
 
![image](https://user-images.githubusercontent.com/47384012/60508256-6991cb00-9cca-11e9-95be-5e13a7a8b97d.png)

Without, it looks like this:
 
![image](https://user-images.githubusercontent.com/47384012/60508266-6e567f00-9cca-11e9-9d6b-39b499e9a2d1.png)

With specific IPython display methods it could e.g. look like this in Jupyter:

![image](https://user-images.githubusercontent.com/47384012/60508279-77dfe700-9cca-11e9-9a59-8a0d8a41b826.png)
![image](https://user-images.githubusercontent.com/47384012/60508290-7c0c0480-9cca-11e9-85e6-9a19502ffcfd.png)

OrtnerMichael:
Dear Alex,

Thank you for the feedback!!
We understand that you address two issues:

1. Is there a particular reason why you would structure the classes as you have shown in "example class". Meaning that you would introduce getter and setter methods ? It is our understanding that this is not standard procedure in Python because you can always access all attributes of all objects anytime.

2. We like the idea that there is a string output instead of reference to the object. But we are not sure if there is sufficient demand to justify adding a feature that can easily be mimicked by any user. We think this is just for console use of the library - is there a special use-case for you that we are not aware of? We would like to know why you want this.

Alexboiboi:
Hi @OrtnerMichael, thanks for taking the time for this request. 

1. It was never my intention to restructure the class you have implemented in your library. I just needed a base class to show an example for the  `__repr__` method. The getter and setter methods were just here because I was experimenting on these methods on my side.

2. For use in the jupyter notebook, it is sometimes just convenient to see the magnet parameters at one glance to verify if all paremeters are defined correctly just by calling the variable. In this case it is just an unecessary feature but nice to have. ;)

Alex-

OrtnerMichael:
We get it. Thank you for the feedback !!!

[...]

PR #200, opened 2019-07-04
Class improvements

Eposs74:
# Related Issues
#188 
# Notes
Improved classes moments, magnets and currents, implemented __repr__ classes

### DIFF
--- a/magpylib/_lib/classes/currents.py
+++ b/magpylib/_lib/classes/currents.py
@@ -208,4 +208,4 @@ class Line(LineCurrent):
 
 
 def __repr__(self):
-    return f"current:  {self.current}, vertices: {self.vertices}, position:  {self.position}, angle:  {self.angle}, axis:  {self.axis}"
+    return "name: {} \n current: {}A \n position: x: {}mm, y: {}mm, z: {}mm \n angle: {}degrees \n axis: {}".format(*self.name, *self.current, *self.position, *self.angle, *self.axis)
--- a/magpylib/_lib/classes/currents.py
+++ b/magpylib/_lib/classes/currents.py
@@ -206,4 +206,4 @@ class Line(LineCurrent):
 
 
 def __repr__(self):
-    return "name: {} \n current: {}A \n position: x: {}mm, y: {}mm, z: {}mm \n angle: {}degrees \n axis: {}".format(*self.name, *self.current, *self.position, *self.angle, *self.axis)
+    return "name: {} \n current: {}A \n position: x: {}mm, y: {}mm, z: {}mm \n angle: {}degrees \n axis: x: {}, y: {}, z: {}".format(*self.name, *self.current, *self.position, *self.angle, *self.axis)
--- a/magpylib/_lib/classes/currents.py
+++ b/magpylib/_lib/classes/currents.py
@@ -127,12 +127,12 @@ class Circular(LineCurrent):
         >>> c
             type: current.Circular 
             current: 2.45  
-            dimension: 3.1469 
+            dimension: d: 3.1469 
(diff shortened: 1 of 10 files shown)


OrtnerMichael:
i like most of this. but please add source type in the __rep__ method. e g. something like "magnet.Box" before printing the attributes. Please also show an example what the output woild look like.

magpylib/_lib/classes/currents.py
```diff
@@ -205,3 +205,7 @@ def getB(self, pos):  # Particular Line current B field calculation. Check RCS f
         rotatedPos = rotateToCS(pos, self)
         return getBField(Bfield_CurrentLine(rotatedPos, self.vertices, self.current),  # The B field
                          self)
+
+
+def __repr__(self):
+    return f"current:  {self.current}, vertices: {self.vertices}, position:  {self.position}, angle:  {self.angle}, axis:  {self.axis}"
```
lucasgcb: Let's remove the vertices from the output for this one. It can get really nasty if it comes out on the console (think of a very long spiral with tons of turns).


These are missing the units for current (A), position (mm) and angle (degrees), and some line breaks 
like the request suggested:

image

magpylib/_lib/classes/magnets.py
```diff
@@ -283,4 +283,4 @@ def getB(self, pos):
 
 
 def __repr__(self):
-    return "name: {} \n magnetization: x: {}mt, y: {}mt, z: {}mt \n dimensions: x: {}mm, y: {}mm, z: {}mm \n position: x: {}mm, y:{}mm, z: {} \n axis: {}".format(*self.name, *self.magnetization, *self.dimension, *self.position, *self.axis)
\ No newline at end of file
+    return "name: {} \n magnetization: x: {}mt, y: {}mt, z: {}mt \n dimensions: x: {}mm, y: {}mm, z: {}mm \n position: x: {}mm, y:{}mm, z: {} \n axis: x: {}, y: {}, z:{}".format(*self.name, *self.magnetization, *self.dimension, *self.position, *self.axis)
```
lucasgcb: Please add the docstring with an example for this method.

Also, change `mt` to `mT` which is [mili Tesla](https://en.wikipedia.org/wiki/Tesla_(unit)) unit, and add a space between the angle and the degree string.

magpylib/_lib/classes/moments.py
```diff
@@ -93,5 +93,22 @@ def getB(self, pos):  # Particular Line current B field calculation. Check RCS f
...
 
-def __repr__(self):
-    return "name: {} \n moment: x: {}mt, y: {}mt, z: {}mt \n position: x: {}mm, y: {}mm, z:{}mm \n angle: {}degrees \n axis: x: {}, y: {}, z: {}".format(*self.name, *self.moment, *self.position, *self.angle, *self.axis)
+    def __repr__(self):
+        """
+        This is for the IPython Console
+        When you call a defined dipole, this method shows you all its components.
+
+        Examples
+    --------
```
lucasgcb: Indentation here and forward is off (should be 8 spaces, this is 4)

lucasgcb: Fix this in the other classes as well.

Eposs74:
Here is the Output of a cylinder per example:
Screenshot Output

Eposs74:
That's the new Output now:

New Output
I just added the missing Unit at Position (z)

OrtnerMichael:
can we make it:

- type: magnet.Cylinder
- Please take dimensions out again.

lucasgcb:
@Eposs74 Could you post a new screenshot of the output? Thanks

lucasgcb:
New output:
image

Issue #197, opened 2019-07-04
Increase test coverage to at least 90%


lucasgcb:
You can also check the coverage on the website:

https://codecov.io/gh/magpylib/magpylib

Locally, you can have an HTML test coverage report in your folder, after running Pytest with the following command:

```
 py.test --junitxml=test-results/magpylib/results.xml --cov-report html --cov-report term:skip-covered --cov=magpylib
```

Add tests that reach the missing lines until the Code coverage reaches 90% or more.

PR #203, opened 2019-07-11
Increased test coverage to 97%

Eposs74:
# Related Issues
#197 
# Notes
Increased test coverage to 97%


lucasgcb:
Have you looked into performing tests with the output of the matplotlib plots for displaySystem?

tests/test_Collection.py
```diff
@@ -490,23 +490,19 @@ def test_GetSweepArray_multiprocessing_error():
         assert isinstance(result,ndarray), type_erMsg
         assert allclose(result,mockResult), errMsg  #check if the field results are the same as the mock results in the array
 
-def test_displaySystem_direct_suppress():
-    from magpylib._lib.classes.magnets import Box
-    b = Box([1,1,1],[1,1,1],pos=(5,5,5))
-
-    a = Collection()
+def test_displaySystem_box():
```
lucasgcb: Rename this to `test_displaySystem` and add a docstring explaining what is expected of this test (what is the behavior being tested?)

tests/test_Collection.py
```diff
@@ -499,4 +499,14 @@ def test_DisplaySystem():
                [.1,.1,.1], # float
                b.position] # float64
 
-    a.displaySystem(markers,suppress=False, direc=True, subplotAx=None)
\ No newline at end of file
+    a.displaySystem(markers,suppress=False, direc=True, subplotAx=None)
```
lucasgcb: Like we discussed in the office, we shouldn't test `suppress=False` because this results in `pyplot.show()` being called and opening windows while tests occur (which makes it a pain to test the library if we run the test suite while working on a feature).

tests/test_base/test_base.py
```diff
@@ -216,3 +216,15 @@ def test_MagMoment_initialization_bad_Mag():
     with pytest.raises(AssertionError):
         MagMoment(moment=badMoment, pos=pos,
                   angle=angle, axis=axis)
+"""
+def test_all_axis_zero():
```
lucasgcb: Remove commented out code.

OrtnerMichael:
Hey @Eposs74,

We've gone through the tests and while they technically increase coverage, they do not exactly test any behavior (there are no assertions, so the content of the plot isn't predicted ).

This PR will stay up while we consider how this can be arranged, but until we can figure out how to test the plots properly (maybe with [numpy.save](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.save.html#numpy.save)?) we'd prefer to keep the coverage we have.

tests/test_Collection.py
```diff
@@ -488,4 +488,23 @@ def test_GetSweepArray_multiprocessing_error():
...
+    s4 = source.current.Circular(2.45, 3 , pos=(5,5,5))
+    s5 = source.current.Line(2.45, [[2,.35,2],[10,2,-4],[4,2,1],[102,2,7]],pos=(5,5,5))
+    s6 = source.moment.Dipole([1,1,1], pos=(5,5,5))
+
+    a = Collection(s1, s2, s3, s4, s5, s6)
+    markers = [[0,0,0],    # int
+               [.1,.1,.1], # float
+               s1.position, s2.position, s3.position, s4.position, s5.position, s6.position] # float64
+
+    a.displaySystem(markers,suppress=True, direc=True, subplotAx=None)
```
OrtnerMichael: Find out how we can save and compare the output of this line within the test.

lucasgcb:
This should be working now, coverage is officially at 97% once this gets in.

PR #222, opened 2019-11-29
Magpylib vector

OrtnerMichael:
Restructuring of the library.

### Changed
- Restructuring
  - displaySystem is now a top-level function, not a Collection method anymore.
  - getBsweep and multiprocessing options have been completely removed, this functionality
    should be overtaken by the new vector functionality which uses numpy native vectorized 
    code paradigm. If mkl library is set (test by numpy.show_config()) numpy will also 
    automatically use multiporcessing. Code parallelization at magpylib level should be done
    by hand.
- Docstrings are adjusted to work better with intellisense. (Problems with *.rst code)
- public rotatePosition() is now called angleAxisRotation(), former private angleAxisRotation
    is now called angleAxisRotation_priv().

### Added
- Performance computation trough vector functionality included in new top-level subpackge "vector"
- Vectorized versions of math functions added to "math" subpackage


OrtnerMichael:
fix version numbering etc first

Issue #224, opened 2019-12-01
Sphinx not running autodoc in 2.0


lucasgcb:
Sphinx is currently not finding the latest docs (2.0.0) with autodoc.
This may be an `__init__.py` issue.

OrtnerMichael:
@lucasgcb 
i noticed that the linter complained on some imports in conf.py

any thoughts ?

PR #225, opened 2019-12-01
 Fix Sphinx autodoc, Py 3.6 support

lucasgcb:
# Related Issues

#224 

# Notes
 
Reintroduced code that was commented out across `__init__.py` files in the project.
Imports were also changed to reflect the new file names (magnet.py -> magnets.py).


lucasgcb:
@OrtnerMichael please take a look. I've changed the `__init__.py` heavily to reflect changes in filenames and reintroduced the namespace helper `__all__` variable to most of them. 
This allows for both Python 3.6 users to use the library and for Sphinx to find our code.

I'm not sure why they were commented out as all the commit messages said "initfix". So, I've reintroduced them. If you had a reason for it, you can decline these changed it and we just get rid of Python 3.6 support.

This is the documentation page for this branch:
https://magpylib.readthedocs.io/en/doc-patch-1/

PR #232, opened 2019-12-27
Vector cylinder

OrtnerMichael:
completed vector code for cylinder

added and vectorized all required sub-function

added tests for everything

Speed gain is ~ x100 compared to classic code

Improved classic code performance by 20%

### DIFF
--- a/magpylib/_lib/getBvector.py
+++ b/magpylib/_lib/getBvector.py
@@ -22,11 +22,12 @@
 # page at https://www.github.com/magpylib/magpylib/issues.
 # -------------------------------------------------------------------------------
 from magpylib._lib.fields.PM_Box_vector import Bfield_BoxV
+from magpylib._lib.fields.PM_Cylinder_vector import Bfield_CylinderV
 from magpylib._lib.fields.PM_Sphere_vector import Bfield_SphereV
 from magpylib._lib.mathLib_vector import QconjV, QrotationV, QmultV, getRotQuatV, angleAxisRotationV_priv
 import numpy as np
 
-def getBv_magnet(type,MAG,DIM,POSo,POSm,ANG=[],AX=[],ANCH=[]):
+def getBv_magnet(type,MAG,DIM,POSo,POSm,ANG=[],AX=[],ANCH=[],Nphi0=50):
     """
     This function is used for performance computation when the magnetic field is
     evaluated multiple times. Use this function only when performing more than ~10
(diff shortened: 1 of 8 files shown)


tests/test_MathVector.py
```diff
@@ -88,4 +88,21 @@ def test_angleAxisRotationV():
...
     for pos,ang,ax,anch,sol in zip(POS,ANG,AXIS,ANCHOR,SOL):
-        assert np.amax(np.abs(angleAxisRotation(pos,ang,ax,anch)-sol))<1e-10, "bad angleAxisRotationV"
\ No newline at end of file
+        assert np.amax(np.abs(angleAxisRotation(pos,ang,ax,anch)-sol))<1e-10, "bad angleAxisRotationV"
+
+
+
+def test_ellipticV():
+    #random input
+    INP = np.random.rand(1000,4)
```
lucasgcb: If this test ever fails because of a certain combination, it will be hard to recover the problematic input.
We could use a pseudo number generator with a fixed seed instead.

```python 


seed = 6969
np.random.seed(seed)
test_input = np.random.rand(1000,4)


# print(test_input) 
# Should always return:
# [[0.90778746 0.02226607 0.42943218 0.30438451]
#  [0.6351356  0.53867276 0.74523766 0.88416302]
#  [0.96144443 0.50252932 0.74019615 0.16491463]
#  ...
#  [0.44344936 0.48382863 0.24196537 0.83127193]
#  [0.16418535 0.45008313 0.48605127 0.92810509]
#  [0.09808894 0.20805508 0.73558164 0.71283222]]
```

OrtnerMichael:
needs some more fixing

magpylib/_lib/getBvector.py
```diff
@@ -63,13 +64,18 @@ def getBv_magnet(type,MAG,DIM,POSo,POSm,ANG=[],AX=[],ANCH=[]):
...
     N = len(MAG)
 
     # set field type
     if type == 'box':
         Bfield = Bfield_BoxV
+    elif type == 'cylinder':
+        Bfield = Bfield_CylinderV
     elif type == 'sphere':
         Bfield = Bfield_SphereV
     else:
```
lucasgcb: Raise a Runtime Warning instead of printing inside of functions. Printing may get drowned out, and users can catch and handle the situation better based on their interpreter settings when detecting a warning.

```python
from warnings import warn
 warn('Bad Type or WIP', RuntimeWarning)
```

magpylib/_lib/getBvector.py
```diff
@@ -63,13 +64,18 @@ def getBv_magnet(type,MAG,DIM,POSo,POSm,ANG=[],AX=[],ANCH=[]):
...
+    Nphi0=50 : integer gives number of iterations used when calculating diametral
+        magnetized cylindrical magnets.
     """
 
     N = len(MAG)
 
     # set field type
     if type == 'box':
         Bfield = Bfield_BoxV
+    elif type == 'cylinder':
```
lucasgcb: If this needs to be checked again later in line 96 then this initial check is redundant.
Function input for cylinder also has different format so it may not be a good idea to share a common reference.

OrtnerMichael:
ty, maily wanted to check out this button :))
mfg michael

Lucas Gabriel Coliado Bandeira <notifications@github.com> schrieb am Fr.,
27. Dez. 2019, 19:59:

PR #252, opened 2020-11-20
Development

OrtnerMichael:
# Related Issues
- doku fixes
- mathLib docstring fixes
- remove getPhi, fix unit tests
- replace elliptic with cel for convention


lucasgcb:
We're moving on with Magpylib 3 so this PR no longer applies.

Issue #259, opened 2021-03-02
CI with mag3 - setup.py


OrtnerMichael:
- [x] find out for which (lower) versions the tests will work
 - numpy 1.20.1
 - scipy 1.6.0 (numpy-scipy dependency)
 - matplotlib 3.3.4
 - py_version v3.7, v3.8, v3.9, v3.10
- [x] magpylib3 -> magpylib naming
- [x] packaging (setup.py)

OrtnerMichael:
working for
python 3.7.10,  3.8.8,  3.9.2
numpy 1.20.1
scipy 1.6.0
matplotlib 3.3.4

might work also with some lower version of numpy, scipy, matplotlib, but why bother supporting this ?

lucasgcb:
> working for
> python 3.7.10, 3.8.8, 3.9.2
> numpy 1.20.1
> scipy 1.6.0
> matplotlib 3.3.4
> 
> @lucasgcb might work also with some lower version of numpy, scipy, matplotlib, any reason we should bother supporting this ?

Ideally, supporting earlier dependencies means it'll be easier to set up for users who may not be up to date. That's pretty much it. 

The risk of doing this would be older versions having features we use becoming deprecated, and having to update the library accordingly.

For now we can keep it with the dependencies we identified, I'll be working on the package script soonish.

OrtnerMichael:
pip install . works now

Issue #258, opened 2021-03-02
CI on mag3 - setting up test pipelines


OrtnerMichael:
py_version v3.7, v3.8, v3.9, v3.10
- [ ] testing pipline
     - [x] windows with azure
     - [x] circle ci (linux?)

lucasgcb:
Python 3.10 is not a stable release yet, so I've not added it to the pipeline.
I could still do it, but it would imply in always testing the latest 3.10 release (which is a bit wonky, as it changes often)

OrtnerMichael:
no dont include it.
Anaconda Standard is now python 3.8

OrtnerMichael:
@lucasgcb can we close this ?

lucasgcb:
> @lucasgcb can we close this ?

Yep, it should have closed when PR #268 was merged.

PR #263, opened 2021-03-06
Magpylib3 Packaging

lucasgcb:
# Related Issues

#259 
#258 

# Notes

- Test files that opened data files required appending of absolute path through `os.path.abspath` so this could work on Docker and on my machines.
- setup.py now requires Scipy and newer versions of Matplotlib and Numpy
- setup.py no longer classifies Python 3.6 as a supported platform
- setup.py no longer classifies magpylib as beta, but software in Production
- CircleCI basic configuration added for tests with Python 3.8 and tox
- Added the previous docs folder for Sphinx 

## Todo:
Check if setup.py file is worth making pep8 compliant
Add tests for more platforms (Python 3.7)


OrtnerMichael:
in commit #fac4f4a you have added all the old documentation.

Most of it is invalid, not just incomplete.
Most of the example.py are going to fail now.

... probably all of them :)

can you just put a big "WIP" at the top and some msg "DOCUMENTATION INVALID FOR NEW VERSION 3" update coming

PR #282, opened 2021-04-15
fix wrong Cylinder faces

jcafhe:
# Related Issues

When using the `display()` function, a Cylinder source is displayed in 3D plot with dimensions that differ from the ones defined. Given the sample below, a cylinder with a diameter of 10 mm and a height of 4mm is displayed as  a cylinder of diameter equals to 10 mm and a height of 20 mm.
```python
import magpylib as magpy
from magpylib.magnet import Cylinder
from magpylib._lib.display.disp_utility import faces_cylinder

src1 = Cylinder(mag=[0, 0, 1000], dim=[10, 4])
magpy.display(src1, show_path=False)
faces = faces_cylinder(src1, False)
```

magpylib_issue_display

# Notes
This comes from the `faces_cylinder()` function, where the z component of each vertex of a face is multiplied by the radius `r`, instead of just x and y components.

Furthermore, the double iteration over the `hs` iterable (`-h2, +h2`) causes the duplication of faces, i.e. the 2 first faces returned by `face_cylinder()` function are identical.

### DIFF
--- a/magpylib/_lib/display/disp_utility.py
+++ b/magpylib/_lib/display/disp_utility.py
@@ -58,14 +58,14 @@ def faces_cylinder(src, show_path):
     hs = np.array([-h2,h2])
     phis = np.linspace(0,2*np.pi,res)
     phis2 = np.roll(np.linspace(0,2*np.pi,res),1)
-    faces = [r*np.array([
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
Looks good, ty for spotting this one !

Issue #298, opened 2021-04-22
[REQUEST] - treat objects with short paths as static


OrtnerMichael:
**problem**
getB doesnt work when objects have different pathlegths.
But objects without a path are treated as static and automatically tiled up.

**feature**
It was requested that objects with shorter paths are automatically considered as static when their paths end.

OrtnerMichael:
closed with pr #307

Issue #287, opened 2021-04-17
Intergrate Circular current class


OrtnerMichael:
Reevaluate code of Circular line current and integrate into v300 code

PR #307, opened 2021-05-19
Magpylib3 development

OrtnerMichael:
# Related Issues
#298 should be solved now - short paths are treated as static
#287 completed with graphic integrstion

# Notes
Level2 layout also massively improved


lucasgcb:
@OrtnerMichael may I delete this branch now that we have retired it?

Issue #292, opened 2021-04-18
Review v3 docstrings


OrtnerMichael:
**all docstings should be in scipy format**

 - only UI level docstrings for v300

OrtnerMichael:
same as #320

PR #321, opened 2021-06-03
Fix Docstring Menu

lucasgcb:
# Related Issues
#292 
# Notes
- The index.rst file had been modified by an earlier commit which caused Sphinx to not import the library docstrings into the side bar; This patch fixes this.

- Fixed some grammar on the page and the ``__init__.py`` bullet list.

- Lock Sphinx to 1.8.2 in the requirements because our configuration is exclusive for that version (currently Sphinx is on 4.0).


OrtnerMichael:
great ! thanks a lot !!!

Issue #353, opened 2021-07-07
Matplotlib -> Mayavi or plotly


OrtnerMichael:
The mayavi mlab package offers proper display of intersections

http://docs.enthought.com/mayavi/mayavi/index.html

This is just a nightmare in Matplotlib

![image](https://user-images.githubusercontent.com/47384012/124747698-7cc37b00-df22-11eb-9e49-ad961a4c3e79.png)

In their help they refer to using Mayavi instead.

So if someone is bored and wants to rewrite the display function :)

OrtnerMichael:
From discussion with @Alexboiboi 

Matplotlib should remain as the standard backend (most common library that everyone has)

Mayavi or plotli backends can/should be activated via kwarg in ``display()`` for ppl who want to have it nice.

lucasgcb:
> From discussion with @Alexboiboi
> 
> Matplotlib should remain as the standard backend (most common library that everyone has)
> 
> Mayavi or plotli backends can/should be activated via kwarg in `display()` for ppl who want to have it nice.

This would involve having Mayavi/Plotli as dependencies in the project alongside matplotlib, even if users don't use them.

A (tricky) alternative is creating magpylib display packages people can download then import, relying on a specific version of magpylib. This would take some thought, though.

PR #388, opened 2021-10-15
Draft PR: Implement an optional plotly plotting backend 

Alexboiboi:
# Related Issues

#353 #394

# Motivation
This pull request aims to bring a new plotting backend with the really nice [plotly](https://plotly.com/python/) library, as it brings better and nicer looking 3D rendering as well a more interactive and feature-rich echosystem with it. Additionally, plotly integrates very well by design in web interfaces such as the jupyter environment and also in online documentation.

# Possible caveats
Since plotly relies on a browser rendering it can be slower than matplotlib to render a lot of different objects at once. By adding a second plottting backend, there will be additonal work on maintenance of the package.

# Tasks checklist

Features:
- [x] Check feature parity with matplotlib
- Support all sources:
  - [x] Cuboid
  - [x] Cylinder
  - [x] Sphere
  - [x] Line
  - [x] Circular
  - [x] Dipole
  - [x] CylinderSegment
- Support other objects:
  - [x] Sensors
  - [x] Markers
  - [x] Path rendering
- Safe guards to avoid user blocked in an endless rendering
  - [x] Animation parameters
  - [x] max frame rate
- Support styling with validation
  - [x] plotly 
  - [x] matplotlib
  - [x] user mesh3d
  - [x] Limit style creation overhead when not necessary
- Code quality:
  - [ ] format (no linter complaints allowed)
  - [ ] tests
- Optional additional features:    
  - [x] path animation
  - ~streamlines 2D~ -> streamlines 2D are not magnetic field lines, do not implement
  - streamtubes
  - surface plot of field components or magnitude

gif1

# Visual comparison: matplotlib/plotly
image
image


OrtnerMichael:
#387 

The new structure works but how do we make the linter recognize the Config members?

image

OrtnerMichael:
@Alexboiboi 

The color transition argument is not working for me.

image

```
import magpylib as magpy

magpy.Config.PLOTTING_BACKEND='plotly'
magpy.Config.NORTH_COLOR = 'red'
magpy.Config.MIDDLE_COLOR = 'green'
magpy.Config.SOUTH_COLOR = 'orange'
magpy.Config.COLOR_TRANSITION = .5

# create two parts of magnet
mag1 = magpy.magnet.Cylinder((1,0,1), (2,2))
mag1.display()

```

OrtnerMichael:
@Alexboiboi 

Pylint is complaining a lot about codestyle.

image

image

image

image

Alexboiboi:
> Pylint is complaining a lot about codestyle.

This is dirty work in progress ^^ nowhere near to be ready for proper review. I'll work on that when the api is stable.

OrtnerMichael:
@Alexboiboi 

when displaying a sensor and all pixel are in a line this happens

image

```
sens = magpy.Sensor(pixel=[(0,0,0), (1,0,0), (2,0,0)])
sens.display()
```

Do you think it makes sense to build a box around all pixel ? I can see advantages and disadvantages here.

OrtnerMichael:
@Alexboiboi 

It would be great if we can set an alpha value for colors (opaqueness / see-through).

We often have to cut out one magnet from another magnet - it would be great if we could see the 2nd magnet inside the first one. Any idea ?

Alexboiboi:
> @Alexboiboi
> 
> The color transition argument is not working for me.
> 
> image
> 
> ```
> import magpylib as magpy
> 
> magpy.Config.PLOTTING_BACKEND='plotly'
> magpy.Config.NORTH_COLOR = 'red'
> magpy.Config.MIDDLE_COLOR = 'green'
> magpy.Config.SOUTH_COLOR = 'orange'
> magpy.Config.COLOR_TRANSITION = .5
> 
> # create two parts of magnet
> mag1 = magpy.magnet.Cylinder((1,0,1), (2,2))
> mag1.display()
> ```

fixed here: df95216e7ff0bbe172ec7b3a05c0cc920604da60

Alexboiboi:
> @Alexboiboi
> 
> when displaying a sensor and all pixel are in a line this happens
> 
> image
> 
> ```
> sens = magpy.Sensor(pixel=[(0,0,0), (1,0,0), (2,0,0)])
> sens.display()
> ```
> 
> Do you think it makes sense to build a box around all pixel ? I can see advantages and disadvantages here.

So far, the display method tries to build a bounding box around the extrema positions of the pixels. So far I chose not to display any particular pixel because when showing lots of pixels + long path combination could lead to rendering performance issue since you would need to render a 3D object for every single pixel. The same issues also applies for rendering a lot of scatter points. I will check what are number of pixels (as tetraeders or cubes) we can render before it becomes unusable. 

I guess, if we show pixels, it is not necessary to show the bounding box...

Alexboiboi:
> the `show_path='animate'` is amazing !!!
> 
> \(ᵔᵕᵔ)/ \(ᵔᵕᵔ)/ \(ᵔᵕᵔ)/ \(ᵔᵕᵔ)/
> 
> How do you set all the animation parameters ?
> 
> `frame_rate` etc cannot be set

You can set them directly in the display function (works for me ;) ):

`mag1.display(show_path='animate', duration=5, frame_rate=5)´

Also the `frame_rate`parameter is not a true frame rate. Since the duration is set and also the maximum path length is also determined in advance, the theoretical frame rate needed is given. This is to prevent the browser to try rendering too many frames per seconds, so it is more a max_frame_rate. Maybe I should rename accordingly.

```python
path_lengths = [obj.position.shape[0] if obj.position.ndim>1 else 0 for obj in objs]
    N = max(path_lengths)
    ds_inds = np.arange(N) # indices
    maxpos = duration*frame_rate
    if N>maxpos:
         # -> apply some downsampling

```

Alexboiboi:
@OrtnerMichael 

What do you think about rewriting the `BaseDisplayRepr ` class as shown below?
This prevents rewriting the entire set of parameters and docstring for the class method to match the `display` function in the `display` top level module.
There is one catch though, that it will allow the method to call other objects to be displayed as the function allows positional arguments, but I see this as a minor problem since all the parameter are named arguments only anyway.

```python
from magpylib._lib.display import display

# ALL METHODS ON INTERFACE
class BaseDisplayRepr:
    """ Provides the display(self) and self.repr methods for all objects

    Properties
    ----------

    Methods
    -------
    - display(self, **kwargs)
    - repr
    """

    display = display

    def __init__(self):
        self._object_type = None

 ........
```

-> Example

```import magpylib as mag4

c1 = mag4.magnet.Cuboid(dimension=(1,2,3), magnetization=(1,1,2))
c2 = mag4.magnet.Cuboid(dimension=(1,2,3), magnetization=(1,1,2)).move((2,0,0))
```

Now calling `c2.display(c1)`would display both c1 and c2

OrtnerMichael:
Daily feedback:

1. markers are not displayed in plotly
2. `size_sensors` doesn't work when there are pixels
3. We should introduce a size_pixel argument - I'm starting a display options discussion #394 just for us to keep track of features and to achieve feature parity
4. `show_path='animate'` doesn't work anymore (assertion error) line 135++ in display.py
5. default zoom level is too close
6. zoom just multiplies the view-window with a factor. This only works when the window goes from negative to positive values. If the window is offset, zoom will offset it further. In the code below the zoomed window doesn't include the magnets anymore.
```
import magpylib as magpy

magpy.Config.PLOTTING_BACKEND='plotly'
magpy.Config.COLOR_TRANSITION = 0.2

mag1 = magpy.magnet.Cuboid((0,0,1), (1,1,1), position=(0,0,3))
mag1.rotate_from_angax(-15, 'x', anchor=(0,0,6))
mag1.rotate_from_angax([1]*30, 'x', anchor=(0,0,6), increment=True)
mag1.display(show_path='animate', zoom=2)
```

OrtnerMichael:
points at master ... nononono

Issue #414, opened 2021-11-24
User 3D display representation


Alexboiboi:
# Context
For the upcoming v4 version we implemented the `style` parameter for all `sources` and `sensors`. The user can attach a custom `mesh` object to an existing representation via `style.mesh3d`.

# Problems:
- `plotly` backend  only
- `mesh3d` object only
- not extendable without breaking the api

# Goal:
We should define a stable and extendable api to allow the user to add or replace an object representation with a single or multiple traces.

# Tasks
- [x] agree on a definitive but extendable api
- [x] implement new concepts
- [x] write tests
- [x] update docs (after merge in docs_v4)

Alexboiboi:
@OrtnerMichael What do you think of this ?

Alexboiboi:
example code

```python
import numpy as np
import magpylib as magpy

cuboid = magpy.magnet.Cuboid(magnetization=(1,0,0), dimension=(3, 3 ,3), position=(10,0,0)).rotate_from_angax([72] * 5, "z", anchor=(0, 0, 0), start=0, increment=True)
cuboid.style.model3d.show = False
cuboid.style.model3d.extra = [
    {
        'backend': 'matplotlib',
        'trace': {
            'type':'plot',
            'xs': np.array([-1, -1,  1,  1, -1, -1, -1, -1, -1,  1,  1,  1,  1, 1,  1, -1])/2, 
            'ys': np.array([-1,  1,  1, -1, -1, -1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1])/2, 
            'zs': np.array([-1, -1, -1, -1, -1,  1,  1, -1,  1,  1, -1,  1,  1, -1,  1,  1])/2,
            'ls': '-',
        },
        'show': True
    },
    {
        'backend': 'plotly',
        'trace': {
            'type': 'scatter3d',
            'x': np.array([-1, -1,  1,  1, -1, -1, -1, -1, -1,  1,  1,  1,  1, 1,  1, -1])/2, 
            'y': np.array([-1,  1,  1, -1, -1, -1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1])/2, 
            'z': np.array([-1, -1, -1, -1, -1,  1,  1, -1,  1,  1, -1,  1,  1, -1,  1,  1])/2,
            'mode':'lines',
        },
        'show': True,
    },
    {
        'backend': 'plotly',
        'trace': {
            'type': 'mesh3d',
[...]

PR #416, opened 2021-11-27
Improve extra user 3d model

Alexboiboi:
# Related Issues
#414

# Notes

### DIFF
--- a/magpylib/_src/default_classes.py
+++ b/magpylib/_src/default_classes.py
@@ -3,10 +3,10 @@ from magpylib._src.default_utils import (
     validate_property_class,
     color_validator,
     get_defaults_dict,
+    SUPPORTED_PLOTTING_BACKENDS
 )
 from magpylib._src.style import DisplayStyle
 
-SUPPORTED_PLOTTING_BACKENDS = ("matplotlib", "plotly")
 
 # ON INTERFACE
 class DefaultConfig(MagicProperties):
--- a/magpylib/_src/default_utils.py
+++ b/magpylib/_src/default_utils.py
@@ -4,6 +4,9 @@ from copy import deepcopy
 import collections.abc
 from magpylib._src.defaults import DEFAULTS
 
+
+SUPPORTED_PLOTTING_BACKENDS = ("matplotlib", "plotly")
+
 MAGPYLIB_FAMILIES = {
     "Line": ("current",),
     "Loop": ("current",),
--- a/magpylib/_src/defaults.py
+++ b/magpylib/_src/defaults.py
@@ -42,7 +42,7 @@ DEFAULTS = {
                 },
                 "description": {"show": True, "text": None},
                 "opacity": 1,
-                "mesh3d": {"show": True, "replace":False, "data": None},
+                "model3d": {"show": True, "extra": []},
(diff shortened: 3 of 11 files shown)


OrtnerMichael:
There is a color problem in Plotly:

with matplotlib backend the `extra` tracecolor is automatically the object color.

image

in Plotly this is not the case
image

generated by the following code
```
import numpy as np
import magpylib as magpy

MOD1 = {
    'backend': 'matplotlib',
    'trace': {
        'type':'plot',
        'xs': np.array([-1, -1,  1,  1, -1, -1, -1, -1, -1,  1,  1,  1,  1, 1,  1, -1])/2, 
        'ys': np.array([-1,  1,  1, -1, -1, -1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1])/2, 
        'zs': np.array([-1, -1, -1, -1, -1,  1,  1, -1,  1,  1, -1,  1,  1, -1,  1,  1])/2,
        'ls': '-',},
    'show': True}
MOD2 = {
        'backend': 'plotly',
        'trace': {
            'type': 'scatter3d',
            'x': np.array([-1, -1,  1,  1, -1, -1, -1, -1, -1,  1,  1,  1,  1, 1,  1, -1])/2, 
            'y': np.array([-1,  1,  1, -1, -1, -1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1])/2, 
            'z': np.array([-1, -1, -1, -1, -1,  1,  1, -1,  1,  1, -1,  1,  1, -1,  1,  1])/2,
            'mode':'lines',
        },
        'show': True,
    }

s1 = magpy.magnet.Sphere((1,2,3),1)
s1.style.model3d.extra=[MOD1, MOD2]
s1.style.color='r'

s2 = magpy.magnet.Sphere((1,2,3),1,(5,0,0))
s2.style.model3d.extra=[MOD1, MOD2]
s2.style.color='g'

s3 = magpy.magnet.Sphere((1,2,3),1,(0,5,0))
s3.style.model3d.extra=[MOD1, MOD2]
s3.style.color='b'

s1.rotate_from_angax(45,'z')

magpy.display(s1,s2,s3, style_magnetization_show=False)
magpy.display(s1,s2,s3, backend='plotly', style_magnetization_show=False)

```

OrtnerMichael:
it would be cool if the input of `src.style.model3d.extra` would accept single dicts like

`src.style.model3d.extra=dict1`

right now it only accepts

`src.style.model3d.extra=[dict1]`

Alexboiboi:
I implemented the "minifeature". The user can provide a dictionary but it will we encapsulated into a list. Later access has to be with `__getitem__`:

example:

```python
model3d.extra = dict() # setting property
print(modeld3d.extra[0])` # access property
```
Should we rename `extra` to `extras` to be more explicit ?

Alexboiboi:
> in Plotly this is not the case

fixed in d6a9ab2e04e9e660b3475f461fb6cf26c768ffce

OrtnerMichael (approved):
Looks good now !!!

`extra` is a very acceptable name - alternatives (`user`, `user_defined`, `external` ) are no good

changing this will not lead to a breaking change !!!

Issue #420, opened 2021-12-01
Combine rotate and rotate_from_angax into one method


Alexboiboi:
Currently, two methods are implemented for rotating objects. The `.rotate` method which takes `scipy.spatial.rotation.Rotation` objects as main argument and the `.rotate_from_angax` which takes `angle` and `axis`  as main arguments. Both methods serve in the end the same goal and could possibly be combined into a single `.rotate` method.

This of course would introduce new breaking changes and should be evaluated before the 4.0 release, unless `.rotate_from_angax` is kept with deprecation warning.

# current implementation
```python
def rotate(self, rotation, anchor=None, start=-1, increment=False):

def rotate_from_angax(self,  angle=None, axis=None, anchor=None, start=-1, increment=False, degrees=True):
```
# proposal

**work in progress in https://github.com/magpylib/magpylib/tree/simplify-rotate**

```python
def rotate(self, *args, angle=None, axis=None, anchor=None, start=-1, increment=False, degrees=True):
"""
positional args:
            can be either:
            - single `scipy.spatial.transform.rotation.Rotation` object
            - both `angle` and `axis` objects as defined below
"""
```

## new behavior :
```python
import magpylib as magpy
from scipy.spatial.transform import Rotation as R
from magpylib._src.exceptions import MagpylibBadUserInput
magpy.defaults.display.backend = 'plotly'

rot = R.from_rotvec([0,0,45], degrees=True)
cube = lambda : magpy.magnet.Cuboid((0,0,1),(1,1,1))

tests ="""
cube().rotate(45,'z').orientation.as_rotvec()
cube().rotate(45, (0,0,1)).orientation.as_rotvec()
cube().rotate(angle=45,axis='z').orientation.as_rotvec()
cube().rotate(rot).orientation.as_rotvec()
cube().rotate(45, axis=(0,0,1)).orientation.as_rotvec()
cube().rotate((0,0,1), angle=45).orientation.as_rotvec()
cube().rotate(rot, angle=45).orientation.as_rotvec()
cube().rotate(rot, axis=(0,0,1)).orientation.as_rotvec()
cube().rotate((0,0,1), 45).orientation.as_rotvec()
cube().rotate(45, (0,0,1), angle=45).orientation.as_rotvec()
cube().rotate(45, (0,0,1), axis=(0,0,1)).orientation.as_rotvec()
"""

should_raise = []
should_not_raise = []
for ind,test in enumerate(tests.strip().split('\n')):
[...]

PR #423, opened 2021-12-06
Simplify rotate

Alexboiboi:
Implements #420


OrtnerMichael:
@Alexboiboi 
**Point of discussion**

How can we implement the `rotation()` method so that
1. scipy Rotation objects can be handed over
2. Angle-Axis rotation is default
3.  full scipy Rotation functionality like `R.from_rotvec()` will be accepted

The original thought was
```
obj.rotate()
obj.rotate.from_angax()
obj.rotate.from_quat()
...
```

It would be great if we could somehow fully integrate the `from_quat()`and `from_rotvec()` in the `obj.rotate()` method.

Alexboiboi:
It is possible to do that but we would need to create a custom Rotation class that would also be callable. I don't think there is a simpler way around this. Here some pseudo code to show how it could look like.

```python
rom scipy.spatial.transform.rotation import Rotation as R

class MagpylibRotation:
    def __init__(self, parent_class):
        self._orientation = R.from_quat([(0, 0, 0, 1)])
        self._parent_class = parent_class
    
    def _rotate(self, rotation, anchor=None, start=-1, increment=False):
        '''like original `BaseGeo` rotate function
        needs an access to `position` attr of parent class
        when self._position shoul be called, call self.parent_class._position instead'''
        print('asdf')
        
    def from_angax(angle, axis, anchor=None, start=-1, increment=False, degrees=True):
        '''like original `BaseGeo` rotate function'''
        
    def from_rotvec(*args, anchor=None, start=-1, increment=False, **kwargs):
        rot = Rotation.from_rotvec(*args, **kwargs)
        self._rotate(rot, anchor=anchor, start=start, increment=increment)
        
    #def from_euler -> same as rotvec 
    #def from_matrix -> same as rotvec
    #def from_mrp -> same as rotvec
    #def from_quat -> same as rotvec
    
    __call__ = _rotate # makes the class callable 

class BaseGeo:
    def __init__(self, position, orientation, style=None):
        self.rotate = MagpylibRotation(parent_class=self)
        
    @property
    def orientation(self):
        return self.rotate._orientation
```

Alexboiboi:
We decided to go with #427 instead

PR #427, opened 2021-12-10
Add all scipy Rotation forms as object methods

Alexboiboi:
# Related Issues
#420
#423

# Note

The plan is to add methods that automatically implement the scipy Rotation package objects with in addition the `anchor` argument and the `path` arguments such as `increment` and `start`.

```
s.rotate()
s.rotate_from_angax()
s.rotate_from_rotvec()
s.rotate_from_quat()
s.rotate_from_euler()
s.rotate_from_rpm()
```

advantages over alternatives:
- rotation class methods `s.rotate.from_angax()` is not consistent because rotation is a method
- just `rotation()` doesn't implement the cool scipy stuff

# Tasks
 
- [x] finalize docstrings
- [x] degrees=True for all methods ?
- [x] write tests for all methods
- [x] add changelog entry


OrtnerMichael (approved):
looks good now

PR #442, opened 2021-12-26
Fixing scene range calculation

### LINKED ISSUE #441: Bad matplotlib zoom window
matplotlib `display()` now doesn't zoom properly when there is a path.

```
import numpy as np
import magpylib as magpy

s = magpy.magnet.Cuboid((0,0,1), (1,1,1))
s.move(np.linspace((0,0,0), (0,10,0), 10))
s.display()
```
Creates this figure

image

Plotly backend is doing the right thing

image

Alexboiboi:
fixes #441 

Just a `-` instead of `+` bug fix.

# Notes

Also fixes a regression introduced by #435 were the `path` argument of the `display` function was conflicting with the `style.path.show` property (only for the `matplotlib` backend

### DIFF
--- a/magpylib/_src/display/display.py
+++ b/magpylib/_src/display/display.py
@@ -152,7 +152,7 @@ def display(
         display_matplotlib(
             *obj_list,
             markers=markers,
-            show_path=path,
+            path=path,
             zoom=zoom,
             axis=canvas,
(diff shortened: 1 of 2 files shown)


OrtnerMichael (approved):
working

Issue #450, opened 2022-01-11
CylinderSegment test failing with numpy==1.22 on Linux CI (np.allclose - close call ?)


Alexboiboi:
Following CylinderSegement test fails when running on CI (Linux) with numpy==1.22

https://github.com/magpylib/magpylib/blob/5dcaa4f2cbcd0caa036b1ab0da5dbac91a87a2f8/tests/test_field_cylinder.py#L232-#L243

current alternative : limit numpy version for testing

Alexboiboi:
closed by #455

PR #453, opened 2022-01-13
Individual path properties, display and animation syntax

Alexboiboi:
# Notes
This PR changes the display syntax by separating the animation and path arguments. It is not a breaking change since the syntax former syntax has not been released yet. The indices along the path at which the object is displayed is now a style property and can be set individually for every object. 

### Notable changes in the display syntax
- [x] all animate wording is replaced by animation
- [x] `path=None`  by default to allow individual object settings
- [x] `path` is now a shortcut to `style_path_show`, both can be used, while `path` takes precedence if both set explicitly
- [x] `path='animate', animate_time=5` becomes `animation=True, animation_time=5` or `animation=5` 
- [x] if `animation` is a number, it takes precedence over `animation` time if both are set explicitly
- [x] so far when `animation` is `True` or a positive `number` all individual path indices settings are ignored.
- [x] remove `path` argument (accessible via `style_path`)
- [x] #451 

### Example
```python
import numpy as np
import magpylib as magpy

# Define sources and sensor
dim = 2
coll = magpy.Collection(
    magpy.magnet.Cuboid(magnetization=(0, 1, 0), dimension=[dim] * 3),
    magpy.magnet.Cylinder(magnetization=(0, 1, 0), dimension=[dim] * 2),
    magpy.magnet.Sphere(magnetization=(0, 1, 0), diameter=dim),
)

ts = np.arange(-0.6, 0.6, 0.2)
sens = magpy.Sensor(position=(0, 0, 2), pixel=[(x, y, 0) for x in ts for y in ts])

# Create paths
pathlen = 50
start_positions = np.array([(1.414, 0, 1), (-1, -1, 1), (-1, 1, 1)])
for pos, src in zip(start_positions, coll):
    src.move(np.linspace(pos, pos*pathlen*0.1, pathlen), start=0)
    src.rotate_from_angax(np.linspace(0., 360., pathlen), 'z', anchor=0, start=0)
sens.move(np.linspace((0.,0.,5.), (0.,0.,-5.), 20), start=0)
sens.style.path.show = 1

# display animation
fig = go.Figure()
magpy.display(
    *coll,
    sens,
    backend="plotly"
)
```
image


Alexboiboi:
Tests are failing because of #450, should be fixed when #455 is merged

OrtnerMichael (approved):
all features are working as outlined

Issue #452, opened 2022-01-12
allow grayscale `'.5'` as argument


OrtnerMichael:
Allow for an easy way for creating gray color shades as matplotlib allows it. 

`style.color` should accept `'0.5'`, `.5` or `0.5` as input

PR #454, opened 2022-01-21
Allow for more color input options

Alexboiboi:
# Related Issues
#452

# Changes
This PR extends the range of possible color inputs for magpylib objects when assigning a color via `style.color`. 

## before 
accepted color inputs:
- short common color names ( eg. `'r'`, `'g'` ... )
- css color names (eg. `'blue'`, `'red'` ...)
- hex string (eg. `#ff00ff` )

## after
accepted color inputs:
- short common color names ( eg. `'r'`, `'g'` ... )
- css color names (eg. `'blue'`, `'red'` ...)
- hex string (eg. `#ff00ff` )
- rgb string (eg. `'rgb(127, 255, 0)'`)
- rgb tuple (eg. `(141,24,230)`)
- number between 0 and 1 (eg. `0.5`) -> interpreted as grey shade, 0 is black, 1 is white
- string number between 0 and 1 (eg. `'.01'`) -> interpreted as grey shade, 0 is black, 1 is white

# Notes
Note that any input that is not a css color or a hex string will be converted into a hex string before assigning it to the style color.
```python
import magpylib as magpy
src = magpy.magnet.Cuboid((0,0,1000),(1,1,1))
src.style.color = 0.5
print(src.style.color)
# [out]: #7f7f7f
```


Alexboiboi:
Tests are failing because of #450

OrtnerMichael (approved):
das ist geil und funktioniert gut !!! :)

Issue #407, opened 2021-11-12
review edgesize setting


OrtnerMichael:
this is another useless feature.

The field expressions should be stabilized in such a way that singularities just happen at corners / edges etc.

Issue #378, opened 2021-08-12
Choose input/output units - make computations dimensionless


OrtnerMichael:
# Feature

**add `defaults` options**
     - `defaults.units.magnetization='mT'`
     - `defaults.units.B = 'mT'`
     - `defaults.units.H = 'kA/m'`
     - `defaults.units.current='A'`
     - `defaults.units.length='mm'`

**make computations dimensionless**
why is this not the case yet ?

OrtnerMichael:
- Must keep B and H !
 - maybe add a default option
     - `defaults.units.magnetization='mT'`
     - `defaults.units.B = 'mT'`
     - `defaults.units.H = 'kA/m'`
     - `defaults.units.current='A'`
     - `defaults.units.length='mm'`

Alexboiboi:
some useful links

https://pint.readthedocs.io/en/stable/

Issue #376, opened 2021-08-09
Direct interface (lib)


OrtnerMichael:
As a result of #374 we will not have simple implementations of the respective papers  but possibly combinations and reformulations to achieve numerical stability.

The naming from #361 and #364 is therefore void - we must think of alternatives - any ideas ?

OrtnerMichael:
input unit = output unit ?

OrtnerMichael:
this is uncritical and will be a new 4.1++ feature.

**reason:**
naming is unclear - straight-forward implementations from papers are often unstable and useless.

PR #467, opened 2022-02-05
Numerical stability v40

OrtnerMichael:
# Related Issues

#407 there is no more edgesize setting. when necessary an edgesize is automatically set to np.close( atol=1e-15, rtol=1e-15)

#378 there is no "unit choosing". all computations are now scale invariant !

#376 lib toplevel package now includes all core field functions

Consider development towards a more separated structure
 - core field functions
 - magpylib interface (objects and position/orientation interface)
 - graphics


Alexboiboi:
I can see you renamed some `cylinder_tile` expression with `cylinder_section`. Is there a good reason not to rename it `cylinder_segment` as in `magnet.CylinderSegment` ?

magpylib/_src/fields/special_cel.py
```diff
@@ -134,36 +134,70 @@ def ellipticPi(x, y):
...
-    C5 = 0.04026699568199808
-    C6 = 0.02642521591631124
-    C7 = 0.01868640268367724
-    C8 = 0.01391747699878044
-    C9 = 0.01076947624905629
-    C10 = 0.008581926385966734
-    return (C2*k2**2 + C3*k2**3 + C4*k2**4 + C5*k2**5
-        + C6*k2**6 + C7*k2**7 + C8*k2**8 + C9*k2**9 + C10*k2**10)
+    errtol = .000001
+    while np.any(abs(g-k) > g*errtol):
```
Alexboiboi, line 155: @OrtnerMichael Are you sure this while loop is safe from being endless?

magpylib/_src/input_checks.py
```diff
@@ -9,6 +9,19 @@
 )
 
 
+def check_field_input(inp, origin):
```
Alexboiboi, line 12: field=True, allowed? Why?

magpylib/_src/fields/special_cel.py
OrtnerMichael: yes - but only Mr. Bulirsch knows the details :)

magpylib/_src/input_checks.py
OrtnerMichael: Doesn't feel right to hand the strings `B` and `H`around library internally. It feels more like a user feature.
This way we get to keep the bool `bh` from level 3 to level 1.

OrtnerMichael:
should now be fine - there is just a merge conflict because of file renamings

Alexboiboi:
> should now be fine - there is just a merge conflict because of file renamings

ready to squash now ;)

PR #468, opened 2022-02-06
`__radd__` method for `BaseGeo` objects

Alexboiboi:
Currently the `__add__` method allows to add up `BaseGeo` object to construct a Magpylib `Collection`. With this PR, the `__radd__` method is added in order to enable the use of the builtin `sum` function.

The `__radd__` method is need because the `sum` functions start the summation by doing `0 + bg_obj` which calls the `__add__` method of the  `int` class which has no knowledge of `bg_obj` and would return an error. If the `__radd__` method is however defined for `bg_obj` the summation will fall back to it.

- Example

```python
import magpylib as magpy

cuboids = [
    magpy.magnet.Cuboid((0, 0, 1), (1, 1, 1), position=(2 * i, 0, 0))
    for i in range(10)
]
coll = sum(cuboids)
coll.show()
```
image


OrtnerMichael (approved):
looks good except 0+src gives a collection ... weird

PR #470, opened 2022-02-08
Basegeo copy method

Alexboiboi:
# Description
This PR adds the `copy` method to the `BaseGeo` class. The method can also take keyword arguments as input, which are transmitted to the new object. 

# Notes
The `copy` method for the `Collection` was only returning a shallow copy. Since we have changed its philosophy, the `Collection` inherits from `BaseGeo` and also its new `copy` method which now returns a deep copy of the object.

# Example
```python
import magpylib as magpy

col = magpy.Collection(
    magpy.magnet.Cuboid(magnetization=(1,0,1), dimension=(8, 4 ,6), position=(0,0,0)),
    magpy.magnet.Cylinder(magnetization=(0,1,0), dimension=(8, 5), position=(-15,0,0)),
    style_name='Collection_01'
)
col2 = col.copy(position=(0,0,10), style=dict(color='red'), style_color='orange')
magpy.show(col, col2, style_magnetization_show=False, backend='plotly')
```

### DIFF
--- a/magpylib/_src/obj_classes/class_BaseGeo.py
+++ b/magpylib/_src/obj_classes/class_BaseGeo.py
@@ -103,8 +103,8 @@ class BaseGeo(BaseTransform):
         if style is not None or kwargs: #avoid style creation cost if not needed
             self.style = self._process_style_kwargs(style=style, **kwargs)
 
-    #@staticmethod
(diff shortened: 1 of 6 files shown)


OrtnerMichael (approved):
tested and looks good

Issue #397, opened 2021-11-09
Parameterviz code refactoring


OrtnerMichael:
reducing style class codes with param pyviz package

Alexboiboi:
```python
import param

class MagicParameterized(param.Parameterized):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
    
    def update(self, **kwargs):
        update_with_nested_dict(self, kwargs)
        return self

def update_with_nested_dict(parameterized, nested_dict):
    # Using `batch_call_watchers` because it has the same underlying
    # mechanism as with `param.update`
    # See https://param.holoviz.org/user_guide/Dependencies_and_Watchers.html?highlight=batch_call#batch-call-watchers
    with param.parameterized.batch_call_watchers(parameterized):
        for pname, value in nested_dict.items():
            if isinstance(getattr(parameterized, pname), param.Parameterized):
                if not isinstance(value, dict):
                    raise ValueError(
                        f'The nested parameterized parameter "{pname}" must be '
                        f'updated with a dictionary, not a {type(value).__name__}.')
                update_with_nested_dict(getattr(parameterized, pname), value)
            else:
                setattr(parameterized, pname, value)

class Description(MagicParameterized):
    show = param.Boolean(default=None, allow_None=True)
    text = param.String()

class BaseStyle(MagicParameterized):
    displayname = param.String(default=None, allow_None=True)
    description = param.ClassSelector(Description, default=Description())
    
new_values = {
    "displayname": "some name",
    "description": {"show":True, "text":"some text"}
}

bs = BaseStyle()

bs.update(displayname='asdf')
bs
```

Alexboiboi:
```python
import panel as pn
import plotly.graph_objects as go
import magpylib as magpy

src = magpy.magnet.Cuboid((1, 0, 1), (8, 4, 6))


def draw_fig(**kwargs):
    style_kwargs = {"style_" + k: v for k, v in kwargs.items()}
    fig = go.Figure()
    magpy.show(src, canvas=fig, **style_kwargs, backend="plotly")
    return fig


i = pn.interact(
    draw_fig,
    magnetization_color_north=pn.widgets.ColorPicker(
        name="magnetization_color_north", value="red"
    ),
    magnetization_color_south=pn.widgets.ColorPicker(
        name="magnetization_color_south", value="green"
    ),
    magnetization_color_mode=pn.widgets.RadioBoxGroup(
        name="magnetization_color_mode",
[...]

PR #473, opened 2022-02-09
`style.name` ➡️  `style.label`

Alexboiboi:
# Related Issues
#397

# Notes
if we want to implement #397, the `name` attribute is already reserved in the `param` library. So we change it beforehand to `label` to avoid api change in the future because of this incompatibility. Anyway we also agreed with @OrtnerMichael, that the term `label` makes more sense.


OrtnerMichael (approved):
works fine !

Issue #476, opened 2022-02-13
possibly unwanted copy behavior


OrtnerMichael:
This code just made me realize that style is automatically initiated whenever you make a copy because you set a label.

```python
obj = magpy.Sensor()
print(f"{obj} with position {obj.position}.")

obj2 = obj.copy(position=(2,6,10))
print(f"{obj2} with position {obj.position}.")
```

This might be unwanted because of the style overhead.
Maybe auto-labeling should only hapen when the original object has a style.
Please comment @Alexboiboi 

**If this is not fixed as proposed, the class_BaseGeo copy() docsting example must be reviewed.**

Alexboiboi:
> auto-labeling should only hapen when the original object has a style.

Sounds like the most meaningful behavior to me. I don't really see another way to avoid the overhead otherwise.

Alexboiboi:
I get the following performance behavior:

## Before
<img width="711" alt="image" src="https://user-images.githubusercontent.com/29252289/153775598-d50cde49-31d2-4bdd-bddb-0ffe781a6856.png">


## After (auto-labeling only if object already has a a style)
<img width="579" alt="image" src="https://user-images.githubusercontent.com/29252289/153775648-f2d5cfe1-4875-4d5f-91f5-6279cfc6a87e.png">

PR #477, opened 2022-02-14
Avoid copy overhead

Alexboiboi:
adresses #476


OrtnerMichael (approved):
code looks good, tests confirm expected performance increase

PR #483, opened 2022-03-02
colorize sensor coordinate arrow tails

Alexboiboi:
# Problem

When they have no pixels, multiple sensors cannot be simply distinguished from each other. This PR solves this issue by colorizing the tail of the coordinate arrows separately.

# Notes
This fix is plotly only and does not include a matplotlib equivalent (yet @OrtnerMichael ?). 
Should this be a specific style option to turn on/off (`style.arrow.colorizetails` ) ?

```python
import numpy as np
import magpylib as magpy

t = np.linspace(-0.1,0.1,5)
pixel  = [[x,y,z] for x in t for y in t for z in t]
s0 = magpy.Sensor(position=(2,0,0), style_label='Sensor1', style_size=4)
s0.style.pixel.size = 0.5
sensors = [s0.copy(position=(i,j,-2*k), pixel=p) for i in range(2) for j in range(2) for k,p in enumerate(((0,0,0), pixel))]
magpy.show(sensors, backend='plotly', renderer='browser')
```

## Before

image

## After

image

### DIFF
--- a/magpylib/_src/display/plotly/plotly_sensor_mesh.py
+++ b/magpylib/_src/display/plotly/plotly_sensor_mesh.py
@@ -152,4 +152,3 @@ def get_sensor_mesh(
                 t.extend(trace[k][i[0]:i[1]])
         trace[k] = np.array(t)
     return trace
-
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
i like

PR #482, opened 2022-02-28
Improve CylinderSegment magnetization matplotlib display

Alexboiboi:
This PR addresses a small display weirdness of CylinderSegment magnetization arrow . When the segment angle (phi1 - phi2) is large the arrow starts at a off-center position. With this PR, off-center arrow start is now using the new barycenter property of the Cylinder Segment. 

**EDIT: do not merge before checking #484**

```python
import magpylib as magpy
import numpy as np

sources = []
for k,sign in enumerate([-1,1]):
    for j,mag in enumerate([(1,0,0),(0,1,0),(0,0,1)]):
        for i, phi in enumerate((90,180,270, 350)):
            src = magpy.magnet.CylinderSegment(
                magnetization=np.array(mag)*sign, 
                dimension=(1,2,1,0,phi),
            position=(-6*i,7*j,15*k))
            sources.append(src)
    
magpy.show(sources, style_opacity=0.1)
```
## Before

image

## After
image

@OrtnerMichael please have a look ;)

Issue #480, opened 2022-02-22
Give CylinderSegment a "barycenter" attribute


OrtnerMichael:
The problem is that the position attribute of the CylinderSegment is always in the center of the full cylinder.
```
import magpylib as magpy
s = magpy.magnet.CylinderSegment((0,0,1000), (1,2,1,0,20))
s.show(markers=[s.position], style_marker_symbol='o', style_marker_size=10)
```
![image](https://user-images.githubusercontent.com/47384012/155230654-4635b23b-70af-4e39-9b4f-4643cb838250.png)

When computing interactions or forces or torques we need the geometric center (should also be the start point of the magnetization arrow). Should we add a method `s.barycenter()` to this class ? Would make these computations much simpler.

Alexboiboi:
closed by #484

PR #484, opened 2022-03-03
CylinderSegment barycenter

Alexboiboi:
Adresses #480 

The CylinderSegment magnetization should maybe use this PR for the arrow start position (see #482)

**Question: Should we cache the call for barycenter to avoid recalculation every time it is called (e.g. using `@lru_cache`)**

```
import magpylib as magpy
import numpy as np

cs = magpy.magnet.CylinderSegment(
    magnetization=(100, 0, 0), dimension=(1, 2, 1, 85, 170), style_opacity=0.5
)

cs.rotate_from_angax([76 * i for i in range(0, 5)], "x", anchor=(0, 0, 5), start=0)

cs.show(
    markers=cs._barycenter.tolist(),
    style_marker_size=5,
    style_path_frames=1,
    backend="plotly",
)
```

image

### DIFF
--- a/tests/test_obj_CylinderSegment.py
+++ b/tests/test_obj_CylinderSegment.py
@@ -1,7 +1,32 @@
+import numpy as np
 import magpylib as magpy
+
 
 def test_repr():
     """ test __repr__
     """
-    pm2 = magpy.magnet.CylinderSegment((1,2,3),(1,2,3,0,90))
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
Looks good,

There is a minor docstring error which makes me realize that I did not review the property docstrings. I will fix this in a subsequent update.

"Output shape is matches position and orientation."

PR #503, opened 2022-03-28
Sensor pixel aggregation

Alexboiboi:
# Related Issues
#496, #283

# Notes

This PR addresses field output aggregation over sensor pixel. It now becomes possible to define an aggregating function (as a string) which can be one of ('min', 'max', 'mean', 'median') and which aggregates the output respectively for each sensor pixels.
The aggregation also make it acceptable to have sensors with different pixel shapes to be evaluated at once, as it becomes compatible with a regular `ndarray` structure.

This PR also solves the perfomance issue with sensor back rotation overhead (#283)

A new `pixel_agg` is added to all `getBH` implementations is is `None` by default.

# Tasks
 - [x] add `pixel_agg` at object `getBH` and `magpy.getBH`
 - [x] update `getBH_level2` to aggregate data
 - [x] make it compatible with heterogeneous pixel shapes
 - [x] update docstrings
 - [x] write tests
 - [x] merge #502 into main
 - [x] close #283
 - [x] update changelog

# Examples

```python
iimport magpylib as magpy
import numpy as np
src1 = magpy.magnet.Cuboid((0,0,1000),(1,1,1)).move([[1,0,0]])
sens1 = magpy.Sensor(position=(0,0,1), pixel=np.zeros((4,5,3)))
sens2 = sens1.copy(position=(0,0,2))
sens3 = sens1.copy(position=(0,0,3))

sources = src1,
sensors = sens1, sens2, sens3
B = magpy.getB(sources, sensors, squeeze=False, pixel_agg=None)
print('expected: (1, 2, 3, 4, 5, 3)', 'out: ', B.shape)
B = magpy.getB(sources, sensors, squeeze=False, pixel_agg='mean')
print('expected: (1, 2, 3, 1, 3)', 'out: ', B.shape)
B = magpy.getB(sources, sensors, squeeze=True, pixel_agg=None)
print('expected: (2, 3, 4, 5, 3)', 'out: ', B.shape)
B = magpy.getB(sources, sensors, squeeze=True, pixel_agg='mean')
print('expected: (2, 3, 3)', 'out: ', B.shape)
```

PR #502, opened 2022-03-23
Nested collections

Alexboiboi:
# Related Issues
#488 

# Notes

This PR implements a new **nested collections** philosophy. 

### BREAKING CHANGES
- A child has a unique parent, and cannot be in multiple collections anymore.
- Adding up objects creates a new Collection -> new x1 + x2 +x3 behavior -> new c += x behavior
- No more Nesting lists/tuples as collection input (single flat list input still allowed)
- remove the `__sub__` method

### Base behavior
A Collection spans a local reference frame and holds object references to that frame. 
- [x] Children can be sensors, sources or other collections. Setters and getters apply only for the base tree level and not the nested one.
- [x] Moving/Rotating the Collection applies to the frame with all objects inside. Relative positions add orientations of nested object is kept, including paths. 
- [x] Duplicate objects are not allowed. **A child as a unique parent**.

### `__add__`  behavior
- [x] Simple rule: **addition always results in super collection for all type of objects**:
 - Example 1: `obj1 + obj2` -> `Collection(obj, obj2)` 
 - Example 2: `obj1 + obj2 + obj3` -> `Collection(Collection(obj, obj2), obj3)` 

### `getBH` behavior:
- [x] `getBH` of a collection gives the total field generated by this collection. It sums up all the sources found in the nested structure.
- [x] `getBH` of a sensor collection gives the field at every sensor found in the flatten structure

### Other features
- [x] `set_children_styles` walks down the nested tree (has recursive argument)
- [x] `remove(obj)` looks for the object in the nested structure (has recursive argument)
- [x] Add a `describe(...)` method for all objects and tree view for nested collections 

### Other tasks
- Adapt plotting backend for nested structure
  - [x] plotly
  - [x] matplotlib
- [x] Check consistency when building collections with `__add__`,  `__radd__` and `__sub__`
- Testing  @OrtnerMichael
  - [x] Fix current tests
  - [x] new behavior tests
- [x] Check if nested behavior is an acceptable overhead @OrtnerMichael 
- [x] Update docstrings
- [x] Go back to simple `__repr__` -> `Collection(id=..., label=...)`
- [ ] Update changelog (specify breaking changes) @OrtnerMichael
- [ ] Fix docs @OrtnerMichael

### Examples

```python
import magpylib as magpy
import numpy as np

magpy.defaults.display.style.magnet.magnetization.show = False

[...]

Issue #492, opened 2022-03-11
Add `field_func` attribute to core sources


OrtnerMichael:
As discussed, Magpylib objects should have the field function stored as an attribute.

This would make it very easy to create custom sources with custom object attribute dependent field functions

This would not break the API.

When this is done also review the CustomSource examples and use the dynamic field function option in the quadrupole example

Alexboiboi:
closed by #540

PR #506, opened 2022-03-30
Custom field func rework

Alexboiboi:
### Motivation
To be more inline with the concept of our core field functions, we decided, for the `CustomSource`, to merge the `field_B_lambda` and `field_H_lambda` into a unique `field_func` property . The new `field_func` must have the first two positional arguments as `field` and `observer`.

### Note
This is not a breaking change, since v4 is not released yet.

### Tasks
- [x] Rewrite `CustomSource` class
- [x] Modify `getBH` at relevant places
- [x] Update docstrings (done but @OrtnerMichael please verify)
- [x] Update docs (done but @OrtnerMichael please verify)
- [x] Agree on how the validation should be:
 - for now only checks if `callable`, args=['field', 'observer'] and if `output.shape==input.shape` with `field='B'` only:
  - Check with `field='B'` AND `field_='H'` ?
  - Force to have both cases or in the latter case should raise `NotImplementedError` ? 
  - Check if `field='B'` and `field='H'` have different outputs?
  - Substantial risk to have a silent error of the kind: `getH(...)` -> still outputs Bfield. Relevant?
- [x] Changelog checked (no changes here) 
- [x] All the core field functions have the firsts two positional args as `field: str, observer: np.ndarray`
- [x] All variables names of the core functions match the `getBH_dict` kwargs (done but @OrtnerMichael please verify)
- [x] allow not-implemented
- [x] #502 merged
- [x] #503 merged

### Related Issues
#492

### DIFF
--- a/docs/_pages/page_01_introduction.md
+++ b/docs/_pages/page_01_introduction.md
@@ -550,7 +550,7 @@ import numpy as np
 import plotly.graph_objects as go
 import magpylib as magpy
 
-# define B-field function for custom source
+# define B/H field function for custom source
 def easter_field(field, observers):
     """ points in z-direction and decays with 1/r^3"""
     dist = np.linalg.norm(observers, axis=1)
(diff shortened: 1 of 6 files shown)


OrtnerMichael (approved):
this looks good, i have added some modifications

PR #509, opened 2022-04-02
Apply doctest

Alexboiboi:
# Description

This PR reworks all docstrings to make sure that they all contain valid example codes.

# Notes
- A new "pytest.ini" has been added to include doctest in the classic testing workflow. This way no more non-working docstring examples can be merged in the future.


OrtnerMichael (approved):
very impressive !!!

PR #511, opened 2022-04-04
adding _all parameters

OrtnerMichael:
# What

This adds `children_all`, `sources_all`, `sensors_all` and `collections_all` as collection parameters that return all respective objects in the tree. Returns some of the previous collection behavior.

```python
import magpylib as magpy

s1 = magpy.Sensor(style_label='s1')
s2 = s1.copy()
s3 = s2.copy()

coll = s1 + s2 + s3
coll.describe(format='label')
```

Collection    
├── Collection
│   ├── s1    
│   └── s2    
└── s3        

```
print([s.style.label for s in coll.sensors])
```
['s3']

```
print([s.style.label for s in coll.sensors_all])
```
['s1', 's2', 's3']

```

### DIFF
--- a/docs/examples/examples_03_collections.md
+++ b/docs/examples/examples_03_collections.md
@@ -137,6 +137,23 @@ and makes it possible to directly reference to a child object by index:
 print(coll[0])
 ```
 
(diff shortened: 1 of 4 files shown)


Alexboiboi (approved):
Looks good ;)

Issue #513, opened 2022-04-06
Computation of a full 360° CylinderSegment with the cut-out method?


Alexboiboi:
### Describe the feature you'd like

As mentioned in #512, there is no good reason not to catch full 360° CylinderSegment case internally and compute it with the cylinder core function with the cut-out method.

Lets implement this

### Notes 
Docs should also be updated (see https://magpylib.readthedocs.io/en/latest/examples/examples_01_complex_forms/#union-operation)

Issue #512, opened 2022-04-04
Discontinuity in fields for magpylib.magnet.CylinderSegment at phi-direction "internal interface"


apdioguardi:
### What is your question?

For full cylinder segments, i.e. phi1+360=phi2, getB and getH return zeros at the "internal interface" in the phi-direction of the cylinder segment. After testing some other cases, like a half cylindrical shell, this seems to be an extension of the numerical stability near interfaces as discussed already in the physics and computation section of the docs. It makes sense then, by extension, that the behavior would be funky near an "internal interface." Maybe it would be useful to add a note to the docstring for CylinderSegment for this special case...

Here's an example:

```python
import numpy as np
import magpylib as magpy

# create a cylindrical shell
r1 = 1
r2 = 2
h = 2
phi1 = 0
phi2 = 360    # also occurs for general phi1+360 =phi2 (I tested just the phi1=90 case though)
# with magnetization
m = np.array([1,1,1])
dims = np.array([r1, r2, h, phi1, phi2])
src = magpy.magnet.CylinderSegment(magnetization=m, dimension=dims)

# one gets discontinuities (all zeros) at the boundary
print( src.getB((1.5, -1e-6, 0.5)) )
print( src.getB((1.5,   0.0, 0.5)) )
print( src.getB((1.5,  1e-6, 0.5)) )
# result:
#[0.52602734 0.8830608  0.68333994]
#[0. 0. 0.] # discontinuity produced
#[0.5260268  0.88306033 0.68334   ]

# everything looks good for the points on the opposite side or if one moves the boundary
print( src.getB((-1.5, -1e-6, 0.5)) )
print( src.getB((-1.5,   0.0, 0.5)) )
print( src.getB((-1.5,  1e-6, 0.5)) )
# result:
#[0.4335992  0.88306026 0.59091234]
#[0.43359947 0.88306056 0.59091237]
#[0.43359974 0.88306086 0.5909124 ]
```

Here is a plot to show that the fields behave nicely outside, but there is a discontinuity of zeros in the x-z plane at y=0 in the above example:

![CylinderSegment_internal_interface_example](https://user-images.githubusercontent.com/2583000/161516115-e2b3b103-8774-4d2d-8dfb-a6e9bc65dec7.PNG)

OrtnerMichael:
hi @apdioguardi 

[...]

PR #514, opened 2022-04-06
Cylinder segment performance

Alexboiboi:
# Related Issues
#512, #513 

# Notes
With this PR, full 360° CylinderSegment case is caught internally and computed with the cylinder core function with the cut-out method. This improves computation time for this special case by almost an order of magnitude.

# Tasks
- [x] catch special 360° case
- [x] update tests
- [ ] update docs (https://magpylib.readthedocs.io/en/latest/examples/examples_01_complex_forms/#union-operation) @OrtnerMichael ?

# Example
```python
from magpylib.magnet import Cylinder, CylinderSegment

# ring from CylinderSegment
ring0 = CylinderSegment(magnetization=(0,0,100), dimension=(2,3,1,0,360))
ring1 = CylinderSegment(magnetization=(0,0,100), dimension=(2,3,1,0,359.999))

%timeit ring0.getB((1,2,3))
%timeit ring1.getB((1,2,3))
```
```log
879 µs ± 10.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
7.13 ms ± 83.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```


OrtnerMichael (approved):
its good

PR #524, opened 2022-04-28
Plotly plotting performance

Alexboiboi:
# Description

When plotting a large number of objects (>500). The plotly plottting backend starts to struggle rendering all the objects. When displaying a large collection of similar objects within a collection it is possible to group and merge the similar traces into a much smaller number of traces. This drastically improves browser rendering performance. In this PR, a method is implemented to group and merge traces based on criteria such as `("legendgroup", "line", "marker", "colorscale", ...)`.
Of Course, this only applies if the object are displayed within a Collection and are not unpacked in the `show` function, since otherwise every object gets it's own legend entry and therefore its own trace, which would prevent merging.

# Open topics
More performance increase could be achieved on the construction of the object on python side though.
Since there is no API change and it is also not really a new feature this could be merged and released as a patch (e.g. 4.0.1) @OrtnerMichael what do you think

- [ ] update changelog (+date) 

# Example

It can still take some time to create and merge the objects (approx 10sec in this case), but the rendering of more than 5000 merged cuboids is almost instantaneous.
image

### DIFF
--- a/magpylib/_src/display/plotly/plotly_display.py
+++ b/magpylib/_src/display/plotly/plotly_display.py
@@ -783,24 +783,25 @@ def group_meshes(*traces):
     """Group and merge mesh traces with similar properties. This drastically improves
     browser rendering performance when displaying a lot of mesh3d objects."""
     mesh_groups = {}
-    keys = ("type", "colorscale", "opacity", "legendgroup")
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
looks good to me

PR #526, opened 2022-05-02
Fix magnetization coloring with mesh grouping (plotly)

Alexboiboi:
The magnetization coloring was failing when magnets are grouped into a collection and have quite different sizes and magnetization. The intensity calculation for the magnetization color scale was not normalized and when merging traces was leading to a bad magnetization representation with plotly. 

This PR fixes this regression introduced by #524. 

```python
import magpylib as magpy
from magpylib.magnet import Cuboid, Cylinder, Sphere

cube = Cuboid(magnetization=(0, 0, 1), dimension=(1, 2, 2))
cylinder = Cylinder(magnetization=(0, 100000, 0), dimension=(1, 1), position=(2,0,0))
sphere = Sphere(magnetization=(0, 1, 1), diameter=1, position=(4,0,0))

coll = magpy.Collection(cube, cylinder, sphere)
magpy.show(coll, backend="plotly")
```

# Before
image

# After
image


OrtnerMichael (approved):
i dont fully get the code but we will find out

PR #532, opened 2022-05-11
Fix copy order bug

### LINKED ISSUE #530: Source collection order bug
### Magpylib version

4.x (Latest)

### What happened?

From @BillyTheKidPhysicist via Email:

... I have found that if I have a collection of sources and iterate through them as “for src in collection”, the order changed mid iteration if inside the loop I do “src.copy()”. I’ve attached code that reproduces this if you are interested. I may be misunderstanding something though, in which case disregard.

### Code example

```Python
from magpylib.magnet import Cuboid
from magpylib import Collection

thing1=Cuboid(magnetization=(1,0,0),position=(1,2,3),dimension=(5,5,5))
thing2=Cuboid(magnetization=(1,0,0),position=(10,2,3),dimension=(5,5,5))
thing3=Cuboid(magnetization=(1,0,0),position=(30,2,3),dimension=(5,5,5))
foo=Collection(thing1,thing2,thing3)
print(foo.sources_all)

#this has the expected order
print('expected--------')
for src in foo:
    print(src)

#That has unexpected order
print('different------')
for src in foo:
    src.copy()
    print(src)

#now this order has changed as well
print(foo.sources_all)
```

### Additional context

_No response_

Alexboiboi:
Fixes #530

### DIFF
--- a/magpylib/_src/obj_classes/class_BaseGeo.py
+++ b/magpylib/_src/obj_classes/class_BaseGeo.py
@@ -338,11 +338,13 @@ class BaseGeo(BaseTransform):
         # pylint: disable=import-outside-toplevel
         from copy import deepcopy
 
-        if not self.parent is None:
-            parent = self.parent
-            self.parent = None
+        # avoid deepcopying the deep dependency upwards the tree structure
+        if self.parent is not None:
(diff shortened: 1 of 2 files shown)


OrtnerMichael (approved):
this looks solid, also tests work fine

PR #540, opened 2022-06-23
Refactor `getBH`

Alexboiboi:
# Notes
This PR drastically refactor and simplify the `getBH` logic in a more declarative way.  This allows to automate some processes and reduces the use of many `if` statements.

This does not change any public API, except that the `getBH_dict` function for "Line" also accepts `vertices`as argument.

# Tasks
 - [x] Clarify the need for internal tiling for the `current.Line` case. (see implementation equivalent for the `magnet.Facets` (see 081618e) (@OrtnerMichael ?)
# Example at `getBH_level1`:
## Before
```python
def getBH_level1(**kwargs: dict) -> np.ndarray:
    """Vectorized field computation

    - applies spatial transformations global CS <-> source CS
    - selects the correct Bfield_XXX function from input

    Args
    ----
    kwargs: dict of shape (N,x) input vectors that describes the computation.

    Returns
    -------
    field: ndarray, shape (N,3)

    """
    # pylint: disable=too-many-statements
    # pylint: disable=too-many-branches

    # base inputs of all sources
    src_type = kwargs["source_type"]
    field = kwargs["field"]  # 'B' or 'H'

    rot = kwargs["orientation"]  # only rotation object allowed as input
    pos = kwargs["position"]
    poso = kwargs["observers"]

    # transform obs_pos into source CS
    pos_rel = poso - pos  # relative position
    pos_rel_rot = rot.apply(pos_rel, inverse=True)  # rotate rel_pos into source CS

    # collect dictionary inputs and compute field
    if src_type == "Cuboid":
        mag = kwargs["magnetization"]
        dim = kwargs["dimension"]
        B = magnet_cuboid_field(field, pos_rel_rot, mag, dim)

    elif src_type == "Cylinder":
        mag = kwargs["magnetization"]
        dim = kwargs["dimension"]
        B = magnet_cylinder_field(field, pos_rel_rot, mag, dim)

    elif src_type == "CylinderSegment":
        mag = kwargs["magnetization"]
        dim = kwargs["dimension"]
        B = magnet_cylinder_segment_field_internal(field, pos_rel_rot, mag, dim)

    elif src_type == "Sphere":
        mag = kwargs["magnetization"]
        dia = kwargs["diameter"]
        B = magnet_sphere_field(field, pos_rel_rot, mag, dia)

    elif src_type == "Dipole":
        moment = kwargs["moment"]
        B = dipole_field(field, pos_rel_rot, moment)

    elif src_type == "Loop":
        current = kwargs["current"]
        dia = kwargs["diameter"]
        B = current_loop_field(field, pos_rel_rot, current, dia)

    elif src_type == "Line":
        current = kwargs["current"]
        if "vertices" in kwargs:
            vertices = kwargs["vertices"]
            B = field_BH_line_from_vert(field, pos_rel_rot, current, vertices)
        else:
            pos_start = kwargs["segment_start"]
            pos_end = kwargs["segment_end"]
            B = current_line_field(field, pos_rel_rot, current, pos_start, pos_end)

    elif src_type == "CustomSource":
        if kwargs.get("field_func", None) is not None:
            B = kwargs["field_func"](field, pos_rel_rot)

    else:
        raise MagpylibInternalError(f'Bad src input type "{src_type}" in level1')

    # transform field back into global CS
    B = rot.apply(B)

    return B
```

## After
```python
FIELD_FUNCTIONS = {
    "Cuboid": magnet_cuboid_field,
    "Cylinder": magnet_cylinder_field,
    "CylinderSegment": magnet_cylinder_segment_field_internal,
    "Sphere": magnet_sphere_field,
    "Dipole": dipole_field,
    "Loop": current_loop_field,
    "Line": current_vertices_field,
}


def getBH_level1(
    *,
    source_type: str,
    position: np.ndarray,
    orientation: np.ndarray,
    observers: np.ndarray,
    **kwargs: dict,
) -> np.ndarray:
    """Vectorized field computation

    - applies spatial transformations global CS <-> source CS
    - selects the correct Bfield_XXX function from input

    Args
    ----
    kwargs: dict of shape (N,x) input vectors that describes the computation.

    Returns
    -------
    field: ndarray, shape (N,3)

    """
    # pylint: disable=too-many-statements
    # pylint: disable=too-many-branches

    # transform obs_pos into source CS
    pos_rel_rot = orientation.apply(observers - position, inverse=True)

    # collect dictionary inputs and compute field
    field_func = FIELD_FUNCTIONS.get(source_type, None)

    if source_type == "CustomSource":
        field = kwargs["field"]
        if kwargs.get("field_func", None) is not None:
            BH = kwargs["field_func"](field, pos_rel_rot)
    elif field_func is not None:
        BH = field_func(observers=pos_rel_rot, **kwargs)
    else:
        raise MagpylibInternalError(f'Bad src input type "{source_type}" in level1')

    # transform field back into global CS
    BH = orientation.apply(BH)

    return BH
```


OrtnerMichael (changes requested):
I like this a lot. This refactoring also includes a new feature:
One can now use the direct interface with `Line` and `vertices`. This was not possible before.

1. This is missing in the docstring in level3
2.  When handing in vertices with different lengths this creates an error

```
import magpylib as magpy

B = magpy.getB(
    'Line',
    [(1,2,3)]*2,
    vertices=[[(0,0,0), (1,1,1)], [(0,0,0), (1,1,1), (2,2,2)]],
    current=1
)
```

OrtnerMichael (commented):
The Direct interface with vertices does not really work - did we say we fix this in this pr ?

For example: This creates an unexpected internal error that should at least have been caught:
```python
import numpy as np
import magpylib as magpy

obs = 

B = magpy.getB(
    "Line",
    observers=[(0,0,0)]*5,
    current=1,
    vertices=np.linspace((0,5,5), (5,5,5), 6)
)
```
when instead it should create 5 outputs, one for each observer because there are 5 observer positions, and a "scalar" current input and a "scalar" vertices input.

- [ ] must be resolved

OrtnerMichael (commented):
Also after giving it some thought - I don't fully understand why we need the @Registered decorator and also its chosen structure

My understanding is, that you
1. register the Magpylib classes in `Registered` with some important properties like `field_func`
2. add some properties to the respective object classes, like `_object_type`

**My criticism:**

1. You add non-protected properties like `family` and `field_func` that are invisible because the linter doesnt see through the decorator. 
2. Then you still also have the protected properties `_family` and `_field_func` ... why ?
3. You dont see these properties when looking at the class code. Would it not be easier/more readable to define them at the class level, and then have the decorator extract them and add them automatically to `Registered` ?

[ ] - discussion on the `Registered` decorator

Alexboiboi:
> The Direct interface with vertices does not really work - did we say we fix this in this pr ?
> 
> For example: This creates an unexpected internal error that should at least have been caught:
> 
> ```python
> import numpy as np
> import magpylib as magpy
> 
> B = magpy.getB(
>     "Line",
>     observers=[(0,0,0)]*5,
>     current=1,
>     vertices=np.linspace((0,5,5), (5,5,5), 6)
> )
> ```
> 
> when instead it should create 5 outputs, one for each observer because there are 5 observer positions, and a "scalar" current input and a "scalar" vertices input.
> 
> * [x]  must be resolved

fixed with https://github.com/magpylib/magpylib/pull/540/commits/af6a813c1f2374347252c0c50af604c89a515671

@OrtnerMichael  I also added a test for this use case but maybe check again if everything is alright now.

Alexboiboi:
> 1. register the Magpylib classes in `Registered` with some important properties like `field_func`
yes

> 2. add some properties to the respective object classes, like `_object_type`
yes
  
> 1. You add non-protected properties like `family` and `field_func` that are invisible because the linter doesnt see through the decorator.
> 2. Then you still also have the protected properties `_family` and `_field_func` ... why ?
These are properties with getters and setters. There is only one `field_func` object, really. It has a getter with the private `_field_func` but no setter, as we don't want the user to change it, only see it. The same is True for `family`.

> 3. You dont see these properties when looking at the class code. Would it not be easier/more readable to define them at the class level, and then have the decorator extract them and add them automatically to `Registered` ?
Also possible but you would then still need to implement setters and getters on each class then, unless you make the attribute only private (like it was for `_object_type`)

An alternative for the `field_func` attribute would be to define a "Master" source class from which every current, magnet or custom sources inherits and then define the `field_func` property there. The same idea could be applied to `family` parameter.

OrtnerMichael (approved):
looks good 👍

PR #552, opened 2022-08-05
Add coupled 3D and 2D subplot functionality

Alexboiboi:
This PR implements a complementary syntax for the `show` function, which allows for creating figures with coupled animated subplots in 3D and 2D.

# Notes
This PR does not introduce breaking API changes, but contains deep internal rework of the display code base.

# Examples
```python
import magpylib as magpy
import numpy as np
from IPython.display import HTML

# define sensor and source
sensor = magpy.Sensor()  # pixel=[(0,0,-.2), (0,0,.2)], style_size=1.5)
cyl1 = magpy.magnet.Cylinder(
    magnetization=(100, 0, 0), dimension=(1, 2), style_label="Cylinder1"
)

# define paths
sensor.position = np.linspace((0, 0, -3), (0, 0, 3), 40)
cyl1.position = (4, 0, 0)
cyl1.rotate_from_angax(angle=np.linspace(0, 300, 40), start=0, axis="z", anchor=0)
cyl2 = cyl1.copy().move((0, 0, 5))

# display system in 3D with dict syntax
magpy.show(
    {"objects": (cyl1, cyl2, sensor), "row":1, "col": 1, "output": "model3d"},
    {"objects": (cyl1, cyl2, sensor), "row":1, "col": 2, "output": ("Bx", "By", "Bz"), "sumup": True},
    backend="plotly",
    animation=True,
    layout_height=600,
)


# display system in 3D with context manager
with magpy.show_context(
    cyl1, cyl2, sensor, backend="plotly", animation=False, layout_height=600
):
    magpy.show(col=1, output="model3d")
    magpy.show(col=2, output=("Bx", "By", "Bz"), sumup=False)

# display system in 3D and return animation object
anim = magpy.show(
    {"objects": (cyl1, cyl2, sensor), "col": 1},
    {"objects": (cyl1, cyl2, sensor), "col": 2, "output": ("Bx", "By", "Bz"), "sumup": True},
    backend="matplotlib",
    animation=True,
    layout_height=600,
    return_animation=True,
)

HTML(anim.to_jshtml())
```
gif1

image
gif1


Alexboiboi:
@OrtnerMichael please have a look when you have spare time. Feedback and change suggestions are welcome ;)

OrtnerMichael:
hi @Alexboiboi 

the output is currently not the same as the one that you show in the figures.

e.g. `output='Bxyz'` does not show all three field components in one graph as in your example.

I play around a little bit - can you explain what the idea of the context manager is here ? simplified syntax ?

Alexboiboi:
> the output is currently not the same as the one that you show in the figures.
> e.g. `output='Bxyz'` does not show all three field components in one graph as in your example.

I have changed the syntax since the pull request comment. `Bxyz` is actually Bx^2+By^2+Bz^2 now. To have all 3 separately you need to write `("Bx", "By", "Bz")`. I have now updated the example above.


> I play around a little bit - can you explain what the idea of the context manager is here ? simplified syntax ?

Basically in simplifies the syntax quite nicely. And for animated coupled subplots it is needed, as the final plot command needs the knowledge of the previous calls within the context manager.

OrtnerMichael:
@Alexboiboi 
**feature request - accept also lists as input to coupled subplot show**

this is accepted
`{"objects": (cyl1, sensor), "row":1, "col": 2, "output": ("B", "Bx", "Bz")},`

this not - but it should be
`{"objects": (cyl1, sensor), "row":1, "col": 2, "output": ["B", "Bx", "Bz"]},`

OrtnerMichael:
@Alexboiboi 
**feature discussion - "output"**
Can you describe what are allowed inputs ? `"Bxyz"` seems to be the same as just `"B"`. In any case squared quantities should be denoted like `"B2"` or `"Bx2"` (because you remarked above that Bxyz is the sum of squares) ...

Would it be possible to hand over simple expressions as strings that include "B", "Bx", "By", "Bz" ... and the same for H ?
So e.g. a "sqrt(Bx+By+B)" ?

I'm not a huge fan of this ... but the coupled subplot functionality is very powerful.

Alexboiboi:
> @Alexboiboi **feature request - accept also lists as input to coupled subplot show**
> 
> this is accepted `{"objects": (cyl1, sensor), "row":1, "col": 2, "output": ("B", "Bx", "Bz")},`
> 
> this not - but it should be `{"objects": (cyl1, sensor), "row":1, "col": 2, "output": ["B", "Bx", "Bz"]},`

fixed in c2201260226f4ce5c03859f25f1e3688f1a53e6e

Alexboiboi:
> @Alexboiboi **feature discussion - "output"** Can you describe what are allowed inputs ? `"Bxyz"` seems to be the same as just `"B"`. In any case squared quantities should be denoted like `"B2"` or `"Bx2"` (because you remarked above that Bxyz is the sum of squares) ...

I made a mistake in the comment above. `Bxyz` is equivalent to `B`,  `Hxz` is the `H` norm in the `xz` plane (`sqrt(Hx**2 + Hz**2`)

> Would it be possible to hand over simple expressions as strings that include "B", "Bx", "By", "Bz" ... and the same for H ? So e.g. a "sqrt(Bx+By+B)" ?
> 
> I'm not a huge fan of this ... but the coupled subplot functionality is very powerful.

I don't really to add even more complexity to the output argument. Parsing this kind of expressions, even simple ones, is tricky and error prone.

Alexboiboi:
> @Alexboiboi **about the context manager**
> 
> At first I didn't like it but I have to say its pretty cool :). Here some feedback:
> 
> 1. `show_context` i don't find a good name for the context manager. I would just use `show` again if thats somehow possible ?

Why is this a bad name? In the very popular library `pandas` there is one called `options_context`. I find it very explicit.
Not sure this is even possible to use `show` both as function and context manager...

> 2. I would also not use `magpy.show` inside the context manager but rather a context manager method like this:
> 
> ```
> with magpy.show(magnet, sensor, backend='plotly') as s:
>    s.show('B', row=1)
>    s.show('model3d', row=2)
>    s.show(('B', 'Bx', 'By'), row=1)
> ```

Not sure about this. One other alternative would be something similar as plotly does to hold updating, which is a bit what we are doing here.
plotly:
```python
fig = go.Figure()
with fig.batch_update():
    fig.add_scatter(...)
    fig.add_scatter(...)
    fig.add_scatter(...)
```

> 3. When using the pyvista backend it just fails - instead I would throw an error with a msg "coupled subplot not supported by pyvista backend"

This should work, have you used the latest updated branch version?

OrtnerMichael:
@Alexboiboi

1. can you please describe what options you have as inputs for `output` ?
2. ok overloading a function to also be a context manager does not work ... but `show_context`  - I just find that name odd, inconveniently long when it should shorten things down ... also not a fan of pandas :)
3. I'm unhappy with calling `magpy.show` inside the context manager to produce 2D graphics UNLESS `magpy.show()` generally gets an `output` kwarg that can produce 2D plots. This would even be a helpful feature when you want to fast-produce a plot. Was this the intent all along ? Dont think I get your plotly comment - maybe we should sit down and talk it out.
4. pyvista works

OrtnerMichael:
We agreed on the following:
1. `magpy.show` has `output` as a general new feature, and should thus be used like that inside the context manager (not `s.show()` using `context as s`)
2. Maybe we can come up with a better name, for example: `multiplot` describes the context manager purpose much better. Lets try to come up with some good names, and discuss again

Alexboiboi:
> 1. can you please describe what options you have as inputs for `output` ?

As already written in the `show` docstring:
```python
output: tuple or string, default="model3d"
        Can be a string or a tuple of strings specifying the plot output type. By default
        `output='model3d'` displays the 3D representations of the objects. If output is a tuple of
        strings it must be a combination of 'B' or 'H' and 'x', 'y' and/or 'z'. When having multiple
        coordinates, the field value is the combined vector length (e.g. `('Bx', 'Hxy', 'Byz')`)
        'Bxy' is equivalent to sqrt(|Bx|^2 + |By|^2). A 2D line plot is then represented
        accordingly if the objects contain at least one source and one sensor.
 ```

> 2. ok overloading a function to also be a context manager does not work ... but `show_context`  - I just find that name odd, inconveniently long when it should shorten things down ... also not a fan of pandas :)

Then make a bunch of suggestions ;) We'll vote


> 3. I'm unhappy with calling `magpy.show` inside the context manager to produce 2D graphics UNLESS `magpy.show()` generally gets an `output` kwarg that can produce 2D plots. This would even be a helpful feature when you want to fast-produce a plot. Was this the intent all along ? Dont think I get your plotly comment - maybe we should sit down and talk it out.

You can call `magpy.show(src, sens, output="Bx")` and you will get a 2D output only.

> 4. pyvista works

ok

**Comment by OrtnerMichael** [comment]
ok ... we agree to the name `show_context`

the programmers cheer - the engineers are sad

**Comment by Alexboiboi** [comment]
Hi @OrtnerMichael,

please also test the animation output argument for pyvista animation

```python
import magpylib as magpy
import numpy as np

# define sensor and source
sensor = magpy.Sensor(pixel=np.linspace((0, 0, -0.2), (0, 0, 0.2), 2), style_size=1.5)
sensor.style.label = "Sensor1"
cyl1 = magpy.magnet.Cylinder(
    magnetization=(100, 0, 0), dimension=(1, 2), style_label="Cylinder1"
)

# define paths
N = 10
sensor.position = np.linspace((0, 0, -3), (0, 0, 3), N)
cyl1.position = (4, 0, 0)
cyl1.rotate_from_angax(angle=np.linspace(0, 300, N), start=0, axis="z", anchor=0)
cyl2 = cyl1.copy().move((0, 0, 5))
objs = cyl1, cyl2, sensor

with magpy.show_context(
    backend="pyvista", animation=1, sumup=True, animation_output="mp4", mp4_quality=10
) as s:
    s.show(*objs, col=1, output=("Bx", "By", "Bz"))  # from contex
    magpy.show(cyl1, col=2)  # directly
```

OrtnerMichael (approved):
looks good, please add remaining issue

PR #586, opened 2023-02-09
Defer Style initialization to when it is needed

Alexboiboi:
# Notes
When working on some routine code where the creation of magpylib objects is very frequent, specifying style attributes cost some significant additional initialization time, so we tend to avoid using them in these cases. However it is quite convenient to be able to give at least a `style.label` to our objects in order to track them more easily later.

This PR partly addresses this issue by deferring the style class tree initialization to when it is really called explicitly (e.g. `obj.style`...). It allows to declare some `style_kwargs` at init level without carrying the time penalty right away. This would be even more relevant if we decide to implement #397.
Here some before/after examples to illustrate the use cases:

image

### DIFF
--- a/magpylib/_src/obj_classes/class_BaseGeo.py
+++ b/magpylib/_src/obj_classes/class_BaseGeo.py
@@ -244,6 +244,7 @@ class BaseGeo(BaseTransform):
         """
         if getattr(self, "_style", None) is None:
             self._style = self._validate_style(self._style_kwargs)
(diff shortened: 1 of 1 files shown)


Alexboiboi:
@OrtnerMichael Please merge this, reviewing should only take a couple of minutes! Thx

OrtnerMichael (approved):
makes a lot of sense to me

Issue #595, opened 2023-03-14
Set a dynamic default backend depending on the current environment


Alexboiboi:
### Scope
As already discussed a few times, there are some arguments about defining a default plotting backend for the magpylib library.  Lets review some pros and cons:

- Matplotlib:
  - Pros:
    - pre-installed with magpylib
    - stand-alone (does not require browser)
    - standard library for the scientific community
  - Cons:
    - 3D Rendering is often really bad in some view angles.
    - Rendering of even few objects is very laggy when rotating/zooming the view by mouse
    - Trisurf-edges remain visible (unless deactivating antialiasing which is really ugly a default 80dpi)
    - Does not render magnetization gradient (yet wip #597 -> bad overlapping layers are event more visible then)
    - No dynamic legend
- Plotly:
  - Pros:
    - Modern look
    - Renders 3D properly
    - Mouse interactivity is very reactive
    - Legend is dynamic (hide/show traces by click)
  - Cons:
    - not pre-installed
    - requires browser (not really an issue in my opinion)
    - when not working within jupyter, sometimes the browser window returns blank page (reported by @OrtnerMichael)
- Pyvista:
  - Pros:
    - Modern look
    - Renders 3D properly
    - Mouse interactivity is reactive (except with some jupyter backends, which are deprecated in favor new single one anyway)
    - library brings some nice meshing features useful for `TriangularMesh` generation. Not really related to plotting though...
  - Cons:
    - not pre-installed
    - heavy dependencies
    - the now jupyter backend ([trame](https://kitware.github.io/trame/index.html)) east a lot of resources (ram, cpu)
    - No dynamic legend
  

I think we should consider setting  a dynamic default backend depending on the current environment. 
 - general case: default="matplotlib"
 - in jupyter or Ipython: set the priority order as  `plotly` > `pyvista` > `matplotlib`, fall back to next available until last which is always available.
 - The new default should then be `backend="auto"` instead of `backend=None` or `backend="matplotlib"`

### Notes
This could almost be considered as a breaking change but it is so minor that I would not care so much about the possible user inconvenience of setting matplotlib explicitly in jupyter notebooks environments.
The docs would probably need some rework too, since almost all plots would default to plotly -> it can slow down the docs page rendering, if many plots are shown, is this desirable ?

@OrtnerMichael Please feel free to chime in to share your thoughts here ;)

Alexboiboi:
fixed by #617

PR #597, opened 2023-03-16
Add Matplotlib magnetization coloring

Alexboiboi:
### Scope
When using the matplotlib library there is no straightforward way to plot color gradients properly in 3D. In this PR, a new set of utilities allows to plot the magnetization orientation over every magnet by slicing and retriangulate the underlying graphical mesh object.

### Examples
This is how it looks like 
- in qt from a script
image
- in jupyter notebook
image

### Caveats
So far it is an _opt-in_ behavior, and we need to decide if we make it the new default for the `matplotlib` backend.
There are a few caveat though that will influence this decision:
- The slicing tends to exacerbate the classical underlying matplotlib issue of bad 3D layering. See example below
image

- Without deactivating ant-aliasing it is impossible to make the triangle edges disappear. And without antialiasing, the plots looks much uglier in the jupyter notebook, since the dpi is set to 80. If we increase dpi, the plots are getting huge.
This is how it looks so far:
#### Antialiasing OFF:
- QT from script:
image

- jupyter
image

#### Antialiasing ON:
- QT from script:
image

- jupyter
image

###Notes
- If we implement #595 , we won't have to focus so much on the jupyter side of the plots. For now, in the docs I  implemented a svg renderer by default, that looks equivalent to the QT output. So that the docs don't look so pixelated.

### Open Tasks
- [x] Make it the new default, or not
- [x] Keep default dpi
- [x] Make docs nicer by enforcing svg output (only for the docs)
- [x]  @OrtnerMichael feedback

Please @OrtnerMichael try out the feature and help us make the decision on how to move forward ;)


OrtnerMichael:
decision:
- coloring default in matplotlib (its crappy no matter what)
- antialiased= True
- keep default dpi

Alexboiboi:
@OrtnerMichael, ready to merge

OrtnerMichael (approved):
looks good ty.
we should think about making pyvista or plotly the default backend

Issue #589, opened 2023-02-18
TriangularMesh - self-intersection check


Alexboiboi:
In the current implementation of the `TriangularMesh` magnet class there  are checks against bad triangle orientation, open mesh and disjoint mesh, but none against self-intersection. This issue is to discuss its implementation.

As this feature is computationally expensive, this should be an opt-in parameter.

Once implemented it could be used as 
```python
import numpy as np
import magpylib as magpy

self_intersecting_mesh3d = {
    "x": [-1.0, -1.0, 1.0, 1.0, -1.0, -1.0, 1.0, 1.0, 0.5],
    "y": [-1.0, 1.0, 1.0, -1.0, -1.0, 1.0, 1.0, -1.0, 0.1],
    "z": [-1.0, -1.0, -1.0, -1.0, 1.0, 1.0, 1.0, 1.0, -3.0],
    "i": [7, 0, 0, 0, 2, 6, 4, 0, 3, 7, 4, 5, 6, 7],
    "j": [0, 7, 1, 2, 1, 2, 5, 5, 2, 2, 5, 6, 7, 4],
    "k": [3, 4, 2, 3, 5, 5, 0, 1, 7, 6, 8, 8, 8, 8],
}
vertices = np.array([v for k, v in self_intersecting_mesh3d.items() if k in "xyz"]).T
triangles = np.array([v for k, v in self_intersecting_mesh3d.items() if k in "ijk"]).T

trimesh = magpy.magnet.TriangularMesh(
    magnetization=(0, 0, 1000),
    vertices=vertices,
    triangles=triangles,
    validate_connected=False,
    reorient_triangles=False,
)
trimesh.style.mesh.intersect.marker.size = 15
trimesh.style.mesh.intersect.line.color = "orange"
trimesh.show(style_opacity=0.5, backend="plotly")
```
![image](https://user-images.githubusercontent.com/29252289/218719534-ecc4b740-f6b6-4875-8eb2-86cd4e970858.png)

## Useful links

https://github.com/pyvista/pyvista/issues/2345

Alexboiboi:
```python
import ipywidgets as ipw
import magpylib as magpy
import numpy as np
import plotly.graph_objects as go
from magpylib._src.display.traces_utility import merge_mesh3d
from magpylib._src.fields.field_BH_trimesh import segments_intersect_triangles

magpy.defaults.display.backend = "plotly"


def update(x=1.2, y=0.9, z=1.1, remove=0):
[...]

Issue #569, opened 2022-11-25
TriangularMesh


OrtnerMichael:
Discussion how the FacetBody class should look like.

TBD

from previous posting:

### The FacetBody class

This class should be quite similar to the TriangularMesh class that @Alexboiboi introduced in the original facet branch, but it is still work in progress and features should still be discussed. The idea is that triangular meshes can be handed over easily, that inside-outside checks will be performed by default ( but can be turned off), and that facet orientation is automatically fixed by default (but can also be turned off).

Parameters: magnetization (the usual) and vertices shape (n,3) and triangles shape (n,3)

It seems to be the common choice in defining meshes through a list of unique vertices, and a second list of indices that define the cells. This is done in pyvista, scipy, pymesh, gmesh, ... . Since we already used vertices in other classes I would use this again for the points, and I really like triangles for the index list.

- pymesh.Mesh uses a list of unique vertices called vertices and a list of face index triplets called faces, supporting triangles and quad faces.
- pyvista.PolyData objects use a list of unique vertices called points and a 1D array [3, 1, 2, 3, 3, 2, 3, 4, ...] called faces where the bold number indicates how many indices define the next face. This enables non-triangular meshes.
- scipy.spatial.Delauny uses a list of vertices called points and a list of face index triplets (like PyMesh) called simplices. The points are not unique.

Alexboiboi:
Current work at https://github.com/magpylib/magpylib/tree/facet_body
Tasks
- [x] Field functions
- [x] graphical implementation
- [ ] class naming `FacetBody` or `TriangularMesh`
- [ ] robust closed body check, if implemented (Euler–Poincaré characteristic is insufficient) 
- [x] Object interface
- [x] check `getBH` (inside/outside)
- [x] optional automatic reorienting facets
- [ ] optional check if closed mesh
- [ ] optional check if self intersect

Other tasks
- [ ] docstrings (+ examples)
- [ ] tests
- [ ] changelog
- [ ] docs with examples

OrtnerMichael:
@Alexboiboi I played a bit around with `TriangularMesh`, started to go through your code and here are a few points that we need to discuss before I start changing/implementing anything

## Which inputs and which names

I think the information that defines the `TriangularMesh()` should be collected in two attributes `points/vertices` shape (n,3) and denoting unique vertices, and in `triangles/facets/faces/simplices` shape (m,3) type int to denote the triangles.
 - I would vote for `vertices` for consistency with other classes.
[...]

PR #598, opened 2023-03-16
Add TriangularMesh magnet

Alexboiboi:
### Notes
- This implements the new `TriangularMesh` magnet class discussed in #569

### Tasks

Tasks
- [x] class naming `TriangularMesh`
- [x] Object interface
- [x] Field functions
- [x] check `getBH` (inside/outside)
  - [x] can be deactivated and set explicitly in `core` for performance (otherwise `auto` by default) 
  - [x] warn if mesh is unchecked and not closed 
- [x] body checks (can be deactivated)
  - [x] automatic facets reorienting (default="warn")
  - [x] check if mesh is open (default="warn") 
  - [x] check if body is disconnected (default="warn")
  -  check if mesh self intersects (default = "skip") - #589
- [x] graphical implementation
  - [x] display `TriangularMesh` objects
  - [x] Rework the style structure: 
    - [x] `TriangularMesh` inherits from the `Magnet` family and its style class, if specified defaults can be different for the TriangularMesh classes (use of composition).
    -  [x] `TriangularMesh` also gets `Orientation` properties, by default `show=False`
  - [x] display triangles orientations via `style.orientation` properties
  - [x] display open edges via `style.mesh.open.show` (+ line and marker properties)
  - [x] display disjoint edges via `style.mesh.disconnected.show` (+ line and marker properties)
  - [x] display disjoint parts by cycling through `style.mesh.disconnected.colorsequence` (default `["red", "blue", "green"]`)
  - display intersecting edges via `style.mesh.selfintersecting.show` (+ line and marker properties) - #589

Other tasks
- [x] tests
- [x] docstrings (+ examples)
- [x] changelog
- [x] docs with examples
- address memory issue (See #600)


ARostekMU:
Hi, thanks for adding this feature, I am looking forward to using it in a new release!

I did some testing and it seems to me that as soon as the mesh is bigger, the magpylib starts eating up enormous amounts of memory.

My script `stl.py` can be found here: https://github.com/ARostekMU/magpylib_test_stl

I tested three STLs. `cube.stl` works just fine, as expected. `random_object.stl` is somewhat more complex and computation takes up to 30 GB of RAM, but it does compute something. However, the mesh is not as fine as I would like it to be. `random_object_finer.stl` has a finer mesh, but the computation eats up all my memory (64 GB) and gets killed.

Is this expected behavior or is there a memory leak somewhere?

Alexboiboi:
Hi @ARostekMU,
Thank you for your feedback ;) 
There is some discussion about this issue here:
- #600

I don't think we will address this issue within the library since there are some simple workarounds that the user can implement (see  #600). This is a fundamental problem on how the field calculation is computed within magpylib and cannot be compared with FEM calculation.

OrtnerMichael:
hello @Alexboiboi 

i am alive

image

I went through the PR ... there is a lot of good stuff happening :). I have a lot of questions though.

First : Did we plan to make `magnet_trimesh_field` a core function ?

I'm not so sure. I think the `TriangularMesh` class functions like a nice gimmick on top of the `triangle_field` as an interface to triangular meshes and all the problems that come with those meshes (connected, oriented, ...). I would not describe it as "superseeding" the `Triangle`, but more like providing such a very practical interface.

What is the argument `in_out` doing in `magnet_trimesh_field` ?
I cannot see this being used anywhere in the library. Also I don't understand the code. Maybe you can explain it to me ?

OrtnerMichael:
**Summary:**
1. `magnet_trimesh_field` will not be a core function because of aforementioned reasons
2. `in_out` started a new discussion #608 
3. `TriangularMesh` does not superseed `Triangle` but provides an interface to common Meshers

OrtnerMichael:
**Problem**
`TriangularMesh` takes the geometric average of all the vertices to calculate the `barycenter`.
This leads to 2 problems:
1. When there is a vertex that is not used for a triangle it can shift the barycenter to a strange position (see plot)
2. The barycenter is not the center of mass, but will be shifted towards positions where there are a lot of vertices

image

As a result the following code generates this strage output

``` python
import numpy as np
import magpylib as magpy
import pyvista

# PROBLEM 1
# when vertices are added that do not contribute to triangles they shift the barycenter
# this also measn that the barycenter is not the center of mass !!!

verts = np.array([(0,0,0), (1,0,0), (0,1,0), (0,0,1), (5,0,0)])
tris =  np.array([(0,1,2), (0,1,3), (1,2,3), (0,2,3)])
trim = magpy.magnet.TriangularMesh(
    magnetization=(0,0,1),
    vertices=verts,
    triangles=tris
)
trim.show()
```


image

OrtnerMichael:
**Question 2**

Why does `TriangularMesh` have its own style properties, like `magnetization`?
It's treated like any magnet, so why do those styles show up ?
image

OrtnerMichael:
**Feature 1**

It would be nice to be able to display a `TriangularMesh` as individual triangles with different colors. This will make its easy for you to spot mesh problems, which is the main purpose of this class. Maybe we could optionally draw the frame of the triangles ?

for example here

image

it is difficult to understand what are the triangle shapes.

Something like `magpy.show(*trimesh)` like you can do it with collections of triangles ... or `trimesh.show(style_mesh_grid_show=True)`

image

OrtnerMichael:
@Alexboiboi 

**to Problem 1**

Weighted centroid is not a solution since you dont have any weights. It should just be the centroid, which is the geometric center.

**to Feature1**

As the `TriangularMesh` class is mostly a mesh integration and analysis tool I think its critical that we are able to show the grid for two reasons: Performance and Precision

Performance and Precision both rely heavily on the triangle finesse. You should be able to see where the mesh is fine and where it is coarse to get a feeling about the two. Specifically flat surfaces are currently a nightmare ... you dont see the triangles, so you don't know if you have a performance issue or if you just screwed up the mesh.

The style `style_mesh_grid_show=True` makes a lot of sense to me

**Feature 2**
`TriangularMesh` method `as_triangles()` that returns a collection of triangles.

There are three good reasons I can see why one would introduce such a feature
1. Doing it by hand is a bit annoying ... few lines of code or not readable
2. We can show individual triangles for mesh analysis - not having to worry too much about realizing the `style_mesh_grid_show` thing which would be nicer but is probably annoying to implement.
3. We can select/remove individual triangles for body construction (computation and precision issues)

I am currently using a transformation to a Triangle Collection to perform mesh analysis ...

OrtnerMichael:
**KEEPING TRACK OF DISCUSSION**

- [ ] Question 1
- [ ] Feature 1
- [ ] Feature 2
- [ ] Problem 1
- [ ] Proposal 1

OrtnerMichael:
**Question 2**
Why call it `from_ConvexHull_points` and not just `from_ConvexHull` ?
I'd prefer the 2nd one

Alexboiboi:
> Weighted centroid is not a solution since you dont have any weights. It should just be the centroid, which is the geometric center.

The weights would be given by the triangles areas

> **to Feature1**
> 
> As the `TriangularMesh` class is mostly a mesh integration and analysis tool I think its critical that we are able to show the grid for two reasons: Performance and Precision

The current triangular mesh styles are `mesh.disjoint` and `mesh.open`, I guess we could add `mesh.all` or `mesh.grid` with similar sub-properties


Feature 2
TriangularMesh method as_triangles() that returns a collection of triangles.

I'd prefer `to_triangles`, which is more explicit in the meaning of an actual transformation.

> **Question 2** Why call it `from_ConvexHull_points` and not just `from_ConvexHull` ?

Ok if you want it. It originally was named `from_points` and with the `triangles="ConvexHull"` argument. Maybe still worth consider this, if there is a different way to build triangles from a point cloud. This would just avoid to add a new `classmethod`.

OrtnerMichael:
**barycenter**
right ... but it sounds quite annoying to compute these weights correctly ... if it can even be done ... so what do propose ? ... i'd go for the geometric center, consider to add a `geocenter` parameter which is also the barycenter for all other sources and be happy...maybe deprecate `barycenter` ?

**all others**
i have added all other comments where we agree to the todo list above. Just to clarify: This is not a todo list for you ! Feel free to do some of the changes, but do not feel obliged :)

Alexboiboi:
> [x] Feature 1: making grid visible
> * [x]  Agreed to add a style option `grid` that makes the grid visible.

implemented in e0af801d06b3b052ccc1db3e026f51d3f200d669

Alexboiboi:
> [x] Feature 2: transform to Triangle Collection
> * [ ]  Agreed to add a `to_TriangleCollection` method that transform the `TriangularMesh` into a collection of triangle objects

Implemented in 0546b1bf516b0c98d590aa92899d5a71633d80e9

OrtnerMichael:
Making grid visible works nicely ... except in Matplotlib ... there you only see outside boundaries.

image

Is this a bug or an issue of the plotting method ?
I thought that Poly3DCollection is used which has the option `set_edgecolor`

OrtnerMichael:
the plot-window in plotly and matplotlib is still using the vertices so the following code with one useless vertex in a TriangularMesh magnet produces this:

image

```python
import magpylib as magpy

verts = ((0,0,0), (1,0,0), (0,1,0), (0,0,1), (10,0,0))
trias = ((0,1,2), (0,1,3), (0,2,3), (1,2,3))

trim = magpy.magnet.TriangularMesh(
    magnetization=(100,200,300),
    vertices=verts,
    triangles=trias
)
trim.show(backend='plotly')
```

OrtnerMichael:
**Problem 2**

Its a strange memory problem: Look at the following code.

```python
duck_pv = pv.read('Duck.stl')
duck_pv.plot()

print(duck_pv.points.shape)
print(duck_pv.faces.shape)

duck = magpy.magnet.TriangularMesh.from_pyvista(
    magnetization=(0,0,1),
    polydata=duck_pv,
    validate_closed=False,
    validate_connected=False,
    reorient_triangles=False
)

duck.show(backend='pyvista', style_magnetization_show=False)
```

When I run this, I can easily import `Duck.stl` into pyvista. `duck.plot` takes less than 1s in pyvista.

I can transform this into a magnet with magpylib, but only if I turn all the checks off. The file has 300k vertices and about 2Mio facets. It is more than 30MB in size.

When I want to plot in matplotlib with Pyvista backend I get a memory error after waiting for 10s.

any thought on this one ?

OrtnerMichael:
**Question 3**

@Alexboiboi  

Also I was reviewing our mesh checks.

There are a bunch of libraries that also do checks like that. E.g. `numpy-stl` has a least a closed and a connected check. Looks like pyvista has `inside` checks, `gmsh` probably has everything.

did you ever review the possibilities provided by other libraries and compare them in terms of performance ?

My thought is that our implementations might be good, but they are probably not as efficient as the ones performed by those other very professional libraries.

Alexboiboi:
> ```python
> import magpylib as magpy
> 
> verts = ((0,0,0), (1,0,0), (0,1,0), (0,0,1), (10,0,0))
> trias = ((0,1,2), (0,1,3), (0,2,3), (1,2,3))
> 
> trim = magpy.magnet.TriangularMesh(
>     magnetization=(100,200,300),
>     vertices=verts,
>     triangles=trias
> )
> trim.show(backend='plotly')
> ```

I was aware of this but I first though it could be an indicator of unused vertices, fixed in c943f2880f800bb7fe8d44a025a6d1db7f4502f5

Alexboiboi:
> Is this a bug or an issue of the plotting method ?
> I thought that Poly3DCollection is used which has the option `set_edgecolor`

This is actually a typical matplotlib problem. I you rotate the object you will actually see all the edges. This is because the generic generation of traces before transmitting to the plotting backend is using regular lines to plot the grid. Changing this would require a lot of rework since not all backends support edge properties integrated into the 3d mesh trace object, as it partially is for matplotlib trisurf. Please make it a separate issue if you really want this ;)

Alexboiboi:
> **Problem 2**
> 
> Its a strange memory problem: Look at the following code.
> 
> ```python
> duck_pv = pv.read('Duck.stl')
> duck_pv.plot()
> 
> print(duck_pv.points.shape)
> print(duck_pv.faces.shape)
> 
> duck = magpy.magnet.TriangularMesh.from_pyvista(
>     magnetization=(0,0,1),
>     polydata=duck_pv,
>     validate_closed=False,
>     validate_connected=False,
>     reorient_triangles=False
> )
> 
> duck.show(backend='pyvista', style_magnetization_show=False)
> ```
> 
> When I run this, I can easily import `Duck.stl` into pyvista. `duck.plot` takes less than 1s in pyvista.
> 
> I can transform this into a magnet with magpylib, but only if I turn all the checks off. The file has 300k vertices and about 2Mio facets. It is more than 30MB in size.
> 
> When I want to plot in magpylib with Pyvista backend I get a memory error after waiting for 10s. For field computation i would have understood ... but for plotting I found it strange, especially since pyvista was doing so well.
> 
> any thought on this one ?

This has to to with the fact that in the current implementation, even if the checks are off when creating the object, they are performed by default when `style_mesh_open_show=True` and `style_mesh_disjoint_show=True`.

As I posted in _https://github.com/magpylib/magpylib/issues/569#issuecomment-1461914270_  it already t takes approx 1 sec for 50k elements and I think it scales quadratically. This approach works pretty well as long as the number of elements is not too crazy high. 

log scale
image

linear scale
image


We'll have to think about how to deal with this situation. I guess it would be possible to have `style_mesh_open_show=False` and `style_mesh_disjoint_show=False` and raise an error if set to true and corresponding validation has been turned off in the object creation. Another solution would be to have a timeout (e.g. 5 secs) if the user tries to plot open or disjoint edges, while plotting and performing the checks, and issue a warning if timeout is reached.

Alexboiboi:
> **Question 3**
> 
> @Alexboiboi
> 
> Also I was reviewing our mesh checks.
> 
> There are a bunch of libraries that also do checks like that. E.g. `numpy-stl` has a least a closed and a connected check. Looks like pyvista has `inside` checks, `gmsh` probably has everything.
> 
> did you ever review the possibilities provided by other libraries and compare them in terms of performance ?
> 
> My thought is that our implementations might be good, but they are probably not as efficient as the ones performed by those other very professional libraries.

Of course it is possible to rely on external libraries to perform mesh validation but I think this was the point of magpylib to provide it out of the box. The experimented user who is more performance oriented, the possibility is already there.

I would again suggest to open an issue specific to this topic. I don't think it is worth delaying the feature even more for these edge cases

OrtnerMichael:
@Alexboiboi 

This solves all existing issues.

The next one:

How about we use the names `vertices` and `faces` instead of `facets` ?

There are several reasons I bring this up:
1. faces seems to be a more common term with meshing libraries
2. The package trimesh makes an excellent distinction.The write "facets are groups of coplanar adjacent faces" - a face can also be a facet, but a facet can consist of multiple faces.
3. When I first dived into the TriangularMesh code, I first looked for faces, not for facets ... it seems to be the more natural term.

- [ ] Proposal: switch to faces instead of facets

OrtnerMichael:
The following code shows how to load a STEP file into trimesh, and from there into magpylib. The compatibility between magpylib and trimesh is already awesome.

```python
import trimesh
import magpylib as magpy
fn = r'xxx.step'
mesh = trimesh.Trimesh(**trimesh.interfaces.gmsh.load_gmsh(fn))

mag = magpy.magnet.TriangularMesh(
    magnetization=(0,0,1),
    vertices=mesh.vertices,
    triangles=mesh.faces,
    validate_closed=False,
    validate_connected=False,
    reorient_triangles=False,
)
mag.show(
    style_mesh_open_show=False,
    style_mesh_disjoint_show=False,
    backend='pyvista',
)
```

image

However, based on the memory discussion #610 we should consider if we make `style_mesh_open_show` and `style_mesh_disjoint_show` default True. I personally think that default should be False.

- [ ] default True or False ?

Also can we stop using the terms opt-in and opt-out ? It always confuses me which is which and they seem to origin from permission marketing and are not so much used for software development ...

OrtnerMichael:
**trimesh**

discussion moved to #613 ... you are probably right that it is not needed. would be an excellent doku example tho.

**arguments of TriangularMesh**

You are right ... my proposal is:
1. `vertices` -> `vertices`
2. `triangles` -> `faces`
3. `facets` -> `mesh` or `triangles`

**show erroneous mesh**

I like best compute on the fly. We can patch in a timeout system and warning system later - I added it to #610

Alexboiboi:
I have changed the the `is_closed`, `is_reoriented`, `is_connected`  back to methods, as they may take a while to compute. Property lookups should actually return values nearly instantly and in the case of many faces it would not be the case. Functions are more fitting and show the user that something needs to be computed.
In the case the check has already been performed, it will only return the result, as a form of caching.
This has to be paid attention to if we intend to make the mesh editable. So far it only happens at initialization time.

OrtnerMichael:
**Also what I find really strange:** When you call `obj.is_closed()` then it does not return the value of `_is_closed`, but it calculates `check_closed(obj)` if `_is_closed` is `None` and sets `_is_closed` to that value ... without asking the user if he wants that. Imagine the user just wants to check what is the status - have I already performed the check ?

What I think would be much more intuitive:

1. We have the property `obj.closed`. It is `None` by default. It has no setter. It can have the values `None`, `True`, `False`
2. In addition we have `obj.closed_info` which is also a property without setter. It is `None` by default.
3. We have the method `obj.validate_closed()`

The method will be called by default at init when `validate_closed=True`. It will set `obj.closed` to either `True` or `False`. In addition, if it is `False` it will store the problem information in `obj.closed_info` and throw a warning (see previous post).

OrtnerMichael:
```python
# class init arguments
# # reorient_faces = True / False (perform operation or not)  - 'skip' / 'ignore' / 'warn' / 'raise'
# # check_connected = True / False (perform operation or not) - 'skip' / 'ignore' / 'warn' / 'raise'
# # check_closed = True / False (perform operation or not)    - 'skip' / 'ignore' / 'warn' / 'raise'


# how and when to warn_raise_ignore ?

# class methods / opertions
# # check_closed(disp='ignore'/'warn'/'raise')
# #    --> if status_closed is None:
# #       --> perform check
# #       --> set status_closed
# #       --> if result is False:
# #          --> set status_closed_data
# #          --> ignore/warn/error: (getB problem, view with style_mesh_closed_show=True)
# #    --> return status_closed (can only be True / False)

# # check_connected(disp='ignore'/'warn'/'raise')
# #    --> if status_connected is None:
# #       --> perform check
# #       --> set status_connected
# #       if result is False:
# #          --> set status_closed_data
# #          --> ignore/warn/error: xxx
# #    --> return status_connected (can only be True / False)

# # reorient_faces(disp='ignore'/'warn'/'raise')
# #    --> if status_reoriented is False
# #       --> if status_closed is None
# #           --> check_closed(disp=disp)
# #       --> if status_closed is False
# #           --> ignore/warn/error:
# #       --> perform reorient


# mesh status class attributes
# # status_closed = True / False / None (default)
# # status_connected = True / False / None (default)
# # status_reoriented = True / False (default)
# # status_closed_data = None (default) / edge pairs
# # status_connected_data = None (default) / list of lists with index triplets
```

OrtnerMichael:
hi @Alexboiboi 

some final discussion points:

1. The style command `style_orientation_xx` feels like it should be `style_mesh_orientation_xx`. I know this comes from `Triangle`, but the user doesn't know that. An ideas ?
2. Duplicate faces are not removed. They are detected as open edges. Do we want that ?
3. Would it not make more sense if status_disconnected_data stores only data if the mesh is actually disconnected ? Talking in terms of memory usage when meshes are large
4. You descided against showing different disconnected parts in different colors. Why ? I kind of liked that feature.

OrtnerMichael (approved):
I think this looks good now 💯. Anything else should be patchfixes.
time to merge

PR #618, opened 2023-06-23
fix show disconnected trimesh

Alexboiboi:
fixes displaying disconnected meshes with plotly. Only discovered in the docs


OrtnerMichael (approved):
alles klar

Issue #614, opened 2023-06-07
Change [unit] to (unit)


OrtnerMichael:
It is a common misconception that square brackets are used to descibe units.

The German [DIN Norm 461](https://dx.doi.org/10.31030/1073187) from 1973 explicitly says, that units must not be put in brackets and further relates to DIN 1313

https://www.beuth.de/de/norm/din-461/710844

Also the Physic Review guidelines state the same.

https://journals.aps.org/authors/axis-labels-and-scales-on-graphs-h18

We should abide by this convention and rewrite our documentation and docstrings accordingly

OrtnerMichael:
check in multiplot ?

PR #626, opened 2023-07-06
Doku rework peter

peter-leitner:
This solves only issue #614 displaying units properly.


peter-leitner:
Please do not delete the branch after merge

PR #631, opened 2023-07-19
Plotly extra traces fix

Alexboiboi:
Fixes small bugs related to the use of extra traces for external use cases.


OrtnerMichael (approved):
i dont fully understand what you did :( but I trust that its ok

PR #634, opened 2023-07-20
Add default plotting description for Collections

Alexboiboi:
This adds a default description to the the Collection objects. This PR also makes the draw function `get_trace` public and therefore easily overridable when subclassing any magpylib core class. Some larger code base has been refactored, mainly in the `traces_generic.py` module.

```python
import magpylib as magpy

sens = [magpy.Sensor(position=(0,0,2*i)) for i in range(5)]
src = [magpy.magnet.Cuboid(magnetization=(0,0,1), dimension=(1,1,1), position=(0,i/5,0)) for i in range(5)]
c1 = magpy.Collection(sens[0], style_label='Sensor Collection', style_color="red")
c2 = magpy.Collection(*src[:3], style_label='Source Collection', style_description="custom description")
c3 = magpy.Collection(*sens[1:], *src[3:])

magpy.show({"objects": c1, "col":1}, {"objects": c2, "col":2}, {"objects": c3, "col":3}, backend="matplotlib", figsize=(20,10))
```

image


Alexboiboi:
> 1. This works only in plotly, not in the other two backends.

Will fix it. If we agree on below


> 2. Also: when did we decide that we can show objects without dimension ? I like the idea - but I don't remember that we discussed this. Is this new in this PR ?

Yes it is new. It came up because of a bug (see below). Then I implemented it


```python
import magpylib as magpy

src = magpy.magnet.Cuboid()
magpy.show(magpy.Collection(src))
```
raises `MagpylibMissingInput: Parameter `dimension` of Cuboid(id=140665705787888) must be set.`

but

```python
import magpylib as magpy

src = magpy.magnet.Cuboid()
magpy.show(magpy.Collection(src))
```
bypasses the `check_function` validation process.
Should we permit it in all cases and display a dimensionless scatter point like in the figure above?

Alexboiboi:
@OrtnerMichael 

I fixed the check_dimension bypass bug.  I'll leave the feature of displaying unset dimension hidden in this PR and I opened an issue (#638) to discuss it.

Ready for review now i think

OrtnerMichael (approved):
seems to work

PR #644, opened 2023-08-01
increase length of face orientation vector

### LINKED ISSUE #636: TriangularMesh bad sign
### Magpylib version

4.x (Latest)

### What happened?

I use `TriangularMesh_from_ConvexHull()` (which btw turns out to be one of the must underestimated features ever) to create a pyramid, and comput the field just below the tip.

image

When i change the pyramid beam-angle slightly, the field changes its sign suddenly. I think the inside-outside check failed, and as a result the the pyramid got oriented badly !

image

### Code example

```Python
import numpy as np
import magpylib as magpy

ts = np.linspace(0,2*np.pi,5)

def points(r0):
    return [(r0*np.cos(t), r0*np.sin(t), 10) for t in ts] + [(0,0,0)]

sens = magpy.Sensor(position=(0,0,-.1))

cone1 = magpy.magnet.TriangularMesh.from_ConvexHull((0,0,1),points(12))
B = print(cone1.getB(sens)[2])

cone2 = magpy.magnet.TriangularMesh.from_ConvexHull((0,0,1),points(13))
B = print(cone2.getB(sens)[2])

sens.show(
    {"objects": cone1, "col":1},
    {"objects":cone2, "col":2},
    style_orientation_show=True, backend='pyvista')
```

### Additional context

_No response_

Alexboiboi:
- Fixes #636 

Also added a test from the issue example to cover it.

### DIFF
--- a/magpylib/_src/fields/field_BH_triangularmesh.py
+++ b/magpylib/_src/fields/field_BH_triangularmesh.py
@@ -162,7 +162,7 @@ def is_facet_inwards(face, faces):
     orient /= np.linalg.norm(orient)  # for single facet numpy is fine
 
     # create a check point by displacing the facet center in facet orientation direction
-    eps = 1e-6  # unfortunately this must be quite a 'large' number :(
+    eps = 1e-5  # unfortunately this must be quite a 'large' number :(
     check_point = face.mean(axis=0) + orient * eps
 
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
yeah what can we do :)

PR #651, opened 2023-08-09
Update field_BH_cylinder_segment.py

fslanovc:
Lower tolerance leads to less errors in the special case treatment


Alexboiboi (approved):
looks good

Issue #676, opened 2023-11-01
Handedness


Nikolai-Hlubek:
Is it possible to change the handedness of a sensor coordinate system? 
I didn't find anything in the manual and it cannot be done by rotation alone. 

I know I can do something like the snippet below, but it is confusing that the scene picture is wrong then. 
```python
def calc_field(sensors):
    # compute field at sensor
    sensors_B = []
    for sensor in sensors:
        B_tmp = sensor.getB(magnet, sumup=True)
        B_tmp[:,2] = (-1)*B_tmp[:,2]  # flip sensor / not possible with rotation
        sensors_B.append(B_tmp)

    return sensors_B
```

OrtnerMichael:
hi @Nikolai-Hlubek,
Sorry for the late reply.

The handedness of the sensor coordinate system is not implemented. But you can easily change your calculation to give the results for a left-handed sensor.

First you have to define in which directions the LH sensor axes should point with respect to the RH sensor that is implemented.
One easy choice is (x,y,z) -> (x,-y,z). In this case the y-axis of the LH sensor simply points into the -y direction of the RH sensor. The computation is achieved by multiplying the output from `getB` with (1,-1,1).

```
getB_LH = getB(...) * np.array((1,-1,1))
```

There is also a fancy way to generate a native magpylib sensor with left-handed system by subclassing the `Sensor` class and returning `getB` as described before. The approach is similar to what we show [here](https://magpylib.readthedocs.io/en/latest/_pages/gallery/gallery_tutorial_custom.html#gallery-tutorial-custom).

Is this what you are looking for?

sincerely,
michael

Nikolai-Hlubek:
Hi Michael

Thank you for the reply. I didn't know that it is not implemented. 
Than you can regard my question as a feature request instead. 

How to change handedness in the result I know - I had some code for this in the question. However, this still means that magpy.show() shows the wrong sensor orientation, which is confusing, if I show the scene to colleagues to discuss results. 

magpylib is really an awesome product.

Best wishes
Nikolai

OrtnerMichael:
hi Nikolai,
Thank you for the praise :).
I understand the problem and we will discuss about a possible implementation and keep this issue open as feature request

Can you give us some context/motivation? Is this for a real sensor? I was under the impression that most sensors should have a right handed coordinate system implemented?

Nikolai-Hlubek:
Hi Michael

The use case is as follows: 
[...]

PR #679, opened 2023-11-04
Fix axis mismatch on show/hide of sensor arrows

Alexboiboi:
This fixes the mismatch of showed or hidden sensor axes. More specifically the "y" and"z" where switched when trying to hide one.

I would suggest to do a patch fix release right away when this is merged.  I'd have a workaround for a Left Handed Sensor which needs this fix (see #676)

```python
import magpylib as magpy

sens = []
for i, k in enumerate("xyz"):
    s = magpy.Sensor(
        position=(i, 0, 0),
        style_label=f"Sensor ({k}-axis hidden)",
        style_size=4,
    )
    s.style.arrows.update({f"{k}_show": False})
    sens.append(s)
magpy.show(sens)
```

### Before
image

### After
image

### DIFF
--- a/magpylib/_src/display/sensor_mesh.py
+++ b/magpylib/_src/display/sensor_mesh.py
@@ -149,7 +149,7 @@ def get_sensor_mesh(
         ]
     )
     indices = ((0, 12), (12, 68), (68, 124), (124, 180))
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
This works as described.
Didnt even know we had this kwarg

PR #682, opened 2023-11-09
Fix deployment release

Alexboiboi:
Fixes Deployment release


OrtnerMichael (approved):
maybe we should switch to git hub flows?

PR #687, opened 2023-11-30
Add optional handedness parameter for Sensors

### LINKED ISSUE #676: Handedness

Alexboiboi:
- fixes #676

This PR adds an optional handedness parameter to the `Sensor` class as discussed in #676. 
I chose to use the `handedness` parameter as `"right"` or `"left"`. When `handedness="left"`, the x-axis is flipped as it would be if would compare the coordinate system with the left hand rule vs right hand rule. Using strings instead of a Boolean gives us the possibility to add some more  options if it is more convenient to flip a another axis (e.g. `"left-y"` or `"left-z"`). The overhead compared to a Boolean check is really minimal  since most of the time only one `Sensor` object is used in `getBH`.

### Example

```python
import magpylib as magpy
import numpy as np

cuboid = magpy.magnet.Cuboid(
    magnetization=(1000, 0, 0),
    dimension=(1, 1, 1),
    position=(0, 1, 0),
)
sens_right = magpy.Sensor(
    position=(-1, 0, 0),
    style_label="Sensor (right-handed)",
    style_sizemode="absolute",
)
sens_left = sens_right.copy(
    position=(1, 0, 0),
    handedness="left",
    style_label="Sensor (left-handed)",
)
sens_coll = magpy.Collection(sens_right, sens_left)
sens_coll.rotate_from_angax(np.linspace(0, 90, 6), "y", start=0)

magpy.show(cuboid, *sens_coll)

print("\nsensor right: \n", cuboid.getB(sens_right).round())
print("\nsensor left: \n", cuboid.getB(sens_left).round())
```

image

image


OrtnerMichael (approved):
looks good, could nat find an error and did not maange to break it.
Didn't know you could use dicts like lists.

Alexboiboi:
> Didn't know you could use dicts like lists.

Could you be more specific?

Alexboiboi:
Hi @Nikolai-Hlubek,

thanks for the review! 

BTW, we just released the latest version of Magpylib (4.5.0) where the handedness feature is included!

Issue #603, opened 2023-04-04
Review Classes naming


Alexboiboi:
Since we always try to have explicit names for all our objects. I think it would make sense to review the current api. 

For the `current.Line`, it would make it clearer when called `current.PolyLine` from the name that the user can generate multiple segments. This may avoid some people unnecessarily creating many `Line` objects.

*Renaming Proposals:*
- Magnets:
  - [x] `Cuboid` -> ok
  - [x] `Cylinder` -> ok
  - [x] `Sphere` -> ok
  - [x] `CylinderSegment` -> ok
  - [x] `Tetrahedron` -> ok
  - [x] `TriangularMesh` -> ok
 - Currents:
   - [x] `Line` -> `Polyline`
   - [x] `Loop` -> `Circle`
- Misc:
  - [x] `Dipole` -> ok
  - [x] `CustomSource` -> ok
  - [x] `Triangle` -> ok
 - Others:
   - [x] `Sensor` -> ok 
 
I would suggest to issue a `DeprecationWarning` for when the old name is used, while still allowing it to avoid breaking changes. We can then remove them when v5 is released (or later).

@OrtnerMichael, What do you think ?

OrtnerMichael:
Hmm ... not sure ...

There a bunch of such "inconsitencies" ... like Loop should be `CircularLoop` and Cylinder the same and `CylinderSegment` should be `CylindricalRingAxialSection` ...

OrtnerMichael:
I like it short and descriptive :)

Alexboiboi:
I think `Ring` is not better since this refers more to an `Annulus`. I would vote for just `Circle`

Alexboiboi:
@OrtnerMichael What is the status about this ?

OrtnerMichael:
Let's do it !!!

`Polyline` and deprecation warning !

Alexboiboi:
Lets review all the names before making any new changes

OrtnerMichael:
I propose `Polyline`

![image](https://github.com/magpylib/magpylib/assets/47384012/c9b3670c-a12b-4ab9-bad2-0375d06681ea)

OrtnerMichael:
But Wikipedia trys to push PolygonalChain on us :)

![image](https://github.com/magpylib/magpylib/assets/47384012/bf7e3d78-4e61-4b98-bc0a-a0f18542465e)

OrtnerMichael:
I think `Polyline` ... or `PolyLine` ?
[...]

PR #690, opened 2023-12-11
Renaming classes: `Line`→`Polyline`, `Loop`→`Circular_Loop`

Alexboiboi:
### Related Issues
- #603

### Notes
As discussed in #603, and to be more explicit with the geometry of the objects.

- `Line` is renamed to `Polyline`
- `current_line_field` renamed to `current_polyline_field`
- `Loop` is renamed to `CircularLoop`
- `current_loop_field` renamed to `current_circular_loop_field`

This PR **does not introduce any breaking changes** but a `DeprecationWarning` are issued whenever the old class name, old core function name or old string argument in `getBH` are used. Output is still generated.
By default, python supresses deprecation warnings so I used a custom `MagpylibDeprecationWarning` to avoid this and so that the user is aware of it. (Pyvista does the same).

raises `MagpylibDeprecationWarning`:
```python
import magpylib as magpy

magpy.current.Loop(...)
magpy.getB("Loop", ....)
magpy.core.current_loop_field(...)

magpy.current.Line(...)
magpy.getB("Line", ....)
magpy.core.current_line_field(...)


OrtnerMichael (approved):
lots of changes ... didn't manage to break it - we will see if there are some bugs :)

PR #706, opened 2023-12-22
Fix Cylinder inside H-field computation

### LINKED ISSUE #703: H-field inside for Cylinder and Cylinder Segment does not match
### Magpylib version

4.x (Latest)

### What happened?

When computing the H-field inside a cylinder, the values do not match the values obtained with an equivalent cylinder segment.

@OrtnerMichael Shall we wait for the v5 release or make a hotfix for v4 already now?

### Code example

```Python
import magpylib as magpy
import numpy as np

field = "H"
obs = np.array([(0.1, 0.2, 0.3)])
pols = np.array([(0, 0, 1)] * len(obs))
d,h = 1,1
dims = np.array([(d, h)]*len(obs))
dims_cs = np.array([(0, d/2, h, 0, 360)]*len(obs))

Bc = magpy.core.magnet_cylinder_field(
    field=field,
    observers=obs,
    dimension=dims,
    magnetization=pols,
)
Bcs = magpy.core.magnet_cylinder_segment_field(
    field=field,
    observers=obs,
    dimension=dims_cs,
    magnetization=pols,
)
print(Bc, Bcs)
# ->output :
#(array([[0.02672707, 0.05345415, 0.50743767]]),
# array([[ 0.02672707,  0.05345415, -0.28833704]]))
```

### Additional context

`Cylinder` field function axial vs transversal mask. seems to be the issue
```python
      if any(mask_ax):  # ax computes B-field
        Bz[mask_tv * mask_inside] -= magz[mask_tv * mask_inside]
    return np.concatenate(((Bx,), (By,), (Bz,)), axis=0).T * 10 / 4 / np.pi
    ```
    ([permalink](https://github.com/magpylib/magpylib/blob/e35b724a8154903d2c6710edb1d254105d884f88/magpylib/_src/fields/field_BH_cylinder.py#L388C5-L389C65))

Alexboiboi:
- fixes #703

also added a test for it, relying on the cylinder segment result.

### DIFF
--- a/magpylib/_src/fields/field_BH_cylinder.py
+++ b/magpylib/_src/fields/field_BH_cylinder.py
@@ -386,7 +386,7 @@ def magnet_cylinder_field(
         return np.concatenate(((Bx,), (By,), (Bz,)), axis=0).T
 
     if any(mask_ax):  # ax computes B-field
-        Bz[mask_tv * mask_inside] -= magz[mask_tv * mask_inside]
+        Bz[mask_ax * mask_inside] -= magz[mask_ax * mask_inside]
     return np.concatenate(((Bx,), (By,), (Bz,)), axis=0).T * 10 / 4 / np.pi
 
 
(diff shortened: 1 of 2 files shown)


Alexboiboi:
@OrtnerMichael Is this ok?

OrtnerMichael (approved):
looks good now

image

Issue #698, opened 2023-12-14
Move to SI units - add polarization


Alexboiboi:
In line with version 5 a new polarization parameter should be added to magnets as well as moving the entire library to SI units.

-  [x] add `polarization` parameter.
  - [x] `magnetization` and `polarization` should be two co-dependent parameters (setting one should set the other automatically)
- [x] Move to SI units - SI unit input should return SI unit output.
  - [x] Inputs 
    - [x] `magnetization`  → _A/m_
    - [ ] `polarization`  → _T_
  - [x] Outputs 
    - [x] `B-field`  → _T_
    - [x] `H-field`  → _A/m_
- [x] Other places of units changes
  - [x] docstrings
  - [ ] input checks
  - [x] graphics
  - [ ] objects `__repr__`
- [x] change argument order for all classes:
     ```python
     def __init__(self, position=(0, 0, 0), orientation=None, <...>, style=None, **kwargs)
     ```
  - [x] Magnets:
    - [x] Cuboid, Cylinder, CylinderSegment
       ```python
        <dimension=None,  polarization=None, magnetization=None>
       ```
    - [x] Sphere
       ```python
        <diameter=None,  polarization=None, magnetization=None>
       ```
    - [x] Tetrahedron, Triangle
       ```python
        <vertices=None,  polarization=None, magnetization=None>
       ```
    - [x] TriangularMesh
       ```python
        <
          position=(0, 0, 0),
          orientation=None,
          vertices=None,
          faces=None,
          polarization=None,
          magnetization=None,
          check_open="warn",
          check_disconnected="warn",
          check_selfintersecting="warn",
          reorient_faces="warn",
       >
       ```
  - [x] Currents
    - [x] Circle
       ```python
       <diameter=None, current=None>
       ```
    - [x] Polyline
       ```python
       <vertices=None, current=None>
       ```
  - [x] Misc
    - [x] Dipole
       ```python
       <moment=None>
       ```
    - [x] CustomSource
       ```python
       <field_func=None>
       ```

OrtnerMichael:
What happening on the core level:

**NEW ARGUMENT ORDERS**
    - field
    - observers
    - position=(0, 0, 0),
    - orientation=None,
    - dimension / diameter / vertices / ...
    - polarization / current / moment
    - magnetization
    - style=None,
    - **kwargs,

**CHANGES IN CORE FUNCTIONS:**
- comment #CORE
- *, no more positionals
- magnetization <> polarizations
[...]

Issue #694, opened 2023-12-11
Throw a warning when magnetization is too far off to avoid silent breaking changes


Alexboiboi:
In order to avoid silent breaking changes in version 5 with the unit changes, we should issue a warning when the magnetization value now in A/m (before mT) is too far off. 

- [x] Define threshold <2000
- [x] Add warning when threshold is reached
- [ ] #724

Alexboiboi:
fixed by #712

PR #712, opened 2023-12-25
Move to SI Units  + Redefine magnetization and add codependency with polarization

OrtnerMichael:
# Related Issues
- #698 
- #694 

# Notes
This PR includes two changes:
- Changing all inputs and outputs to SI units!
- Adding the `polarization` attribute to magnets which is co-dependent with the `magnetization` attribute

**BREAKING CHANGES** - this will be part of Magpylib v5


Alexboiboi:
@OrtnerMichael What about the places where we have degrees. With the new SI units should we also move to rad instead of deg  (see also #718)

Alexboiboi (approved):
Looks fine for now, docs will be updated later

Issue #608, opened 2023-05-24
Add `getM()` and `getJ()` - separate inside/outside checks from core functions


OrtnerMichael:
Core functions are trimmed to performance, the mask generation for inside-outside-edge checks take a lot of time. In most cases users know whether they are inside or outside. One more use of it is to refactor the code to enable the calculation of `getM` and `getJ`.

- [x] can inside-outside checks be separated from core computation ?
- [x] can they be turned on and off in a practical manner ?
- [x] Add `getM()`
- [x] Add `getJ()`

> [!NOTE]  
> `in_out` is currently an unused option in `magnet_trimesh_field`

### to clarify:

- #703

- `Circle`, `Polyline`, `Triangle`, `Dipole` -> what are `getM` and `getJ` outputs ? 
- `Sphere` -> cannot use a generic `convert_BHMJ` because of `polarization/3` factor. What about this factor when calculating  `getM` and `getJ`

OrtnerMichael:
- moving to #647

Alexboiboi:
Reopening this as not addressed. Can still be tracked in  #647

PR #717, opened 2024-01-07
Add `getM` and `getJ` + `in_out`  inside/ouside parameter

Alexboiboi:
### Related Issues
- Implements #608
- Needs #712

### Notes
This PR implements the `getM` and `getJ` respectively for the `magnetization` and the `polarization` similarly to `getB` and `getH`. 

The inside/outside checks are now also separated from the field calculation of magnets. By introducing the `in_out` parameter to the core functions as well as for all `getH`, `getB`, `getM`, `getJ` functions it allows the user to specify the location of the observers relative to the magnet body, affecting the calculation of the magnetic field and optimizing performance. This is especially relevant for `Triangularmesh` for which this check is especially computationally intensive. The following options are in use:
- `'auto'` (default): The location (inside or outside the cuboid) is determined automatically for each observer.
- `'inside'`: All observers are considered to be inside the cuboid; use this for performance optimization if applicable.
- `'outside'`: All observers are considered to be outside the cuboid; use this for performance optimization if applicable.
Choosing 'auto' is fail-safe but may be computationally intensive if the mix of observer locations is unknown.

The following code is for performance comparison of the parameter combinations for each magnet type

```python
from time import perf_counter

import magpylib as magpy
import numpy as np
import pandas as pd
import plotly.express as px

objs_params = {
    "Cuboid": {
        "dimension": [1, 1, 1],
        "func": magpy.core.magnet_cuboid_field,
    },
    "Cylinder": {
        "dimension": [1, 1],
        "func": magpy.core.magnet_cylinder_field,
    },
    "Sphere": {
        "diameter": 1,
        "func": magpy.core.magnet_sphere_field,
    },
    "CylinderSegment": {
        "dimension": [0.25, 0.5, 1, 120, 420],
        "func": magpy.core.magnet_cylinder_segment_field,
    },
    "Tetrahedron": {
        "vertices": [(-1, 0, 0), (1, -1, 0), (1, 1, 0), (0, 0, 1)],
        "func": magpy.core.magnet_tetrahedron_field,
    },
}


def get_time(obj_type, field, in_out, N):
    kw = objs_params[obj_type].copy()
    func = kw.pop("func")
    kw["polarization"] = [1, 1, 1]
    kw = {k: np.repeat([v], N, axis=0) for k, v in kw.items()}
    kw["observers"] = np.linspace((-1, 0, 0), (1, 0, 0), N)
    if in_out is not None:
        kw["in_out"] = in_out
    tic = perf_counter()
    func(field=field, **kw)
    return perf_counter() - tic


def get_time_avg(obj_type, field, in_out, N, max_time=.1):
    t = get_time(obj_type, field, in_out, N)
    runs = 1
    if t > max_time:
        t, runs
    runs = max(2, int(max_time / t))
    t = np.mean([get_time(obj_type, field, in_out, N) for _ in range(runs)])
    return t, runs

def calc_res(in_outs):
    res = []
    for in_out in in_outs:
        for num_of_observers in [1, 10, 100, 1000, 10000]:
            for field in "BH":
                for obj_type in objs_params:
                    t, runs = get_time_avg(obj_type, field, in_out, num_of_observers)
                    d = {
                        "obj_type": obj_type,
                        "num_of_observers": num_of_observers,
                        "field": field,
                        "time": t,
                        "runs": runs,
                        "in_out": in_out,
                    }
                    print(f"{str(d):1000}", end="\r")
                    res.append(d)
    df = pd.DataFrame(res)
    df.to_csv(f"res_{'_'.join(str(v) for v in in_outs)}.csv", index=False)

# calc_res(in_outs=[None]) # magnetization-polarization branch

# calc_res(in_outs = ["auto", "inside", "outside"]) # inside-outside-getM-getJ

df1 = pd.read_csv("res_None.csv")
df2 = pd.read_csv("res_auto_inside_outside.csv")
df = pd.concat([df1, df2])
df["in_out"] = df["in_out"].astype(str)

px.line(
    df,
    x="num_of_observers",
    y="time",
    facet_col="obj_type",
    facet_row="field",
    color="in_out",
    line_dash="in_out",
    log_x=True,
    log_y=True,
    height=600
)
```

image


Alexboiboi:
The following code is for performance comparison of the parameter combinations for each magnet type


```python
from time import perf_counter

import magpylib as magpy
import numpy as np
import pandas as pd
import plotly.express as px

objs_params = {
    "Cuboid": {
        "dimension": [1, 1, 1],
        "func": magpy.core.magnet_cuboid_field,
    },
    "Cylinder": {
        "dimension": [1, 1],
        "func": magpy.core.magnet_cylinder_field,
    },
    "Sphere": {
        "diameter": 1,
        "func": magpy.core.magnet_sphere_field,
    },
    "CylinderSegment": {
        "dimension": [0.25, 0.5, 1, 120, 420],
        "func": magpy.core.magnet_cylinder_segment_field,
    },
    "Tetrahedron": {
        "vertices": [(-1, 0, 0), (1, -1, 0), (1, 1, 0), (0, 0, 1)],
        "func": magpy.core.magnet_tetrahedron_field,
    },
}


def get_time(obj_type, field, in_out, N):
    kw = objs_params[obj_type].copy()
    func = kw.pop("func")
    kw["polarization"] = [1, 1, 1]
    kw = {k: np.repeat([v], N, axis=0) for k, v in kw.items()}
    kw["observers"] = np.linspace((-1, 0, 0), (1, 0, 0), N)
    if in_out is not None:
        kw["in_out"] = in_out
    tic = perf_counter()
    func(field=field, **kw)
    return perf_counter() - tic


def get_time_avg(obj_type, field, in_out, N, max_time=.1):
    t = get_time(obj_type, field, in_out, N)
    runs = 1
    if t > max_time:
        t, runs
    runs = max(2, int(max_time / t))
    t = np.mean([get_time(obj_type, field, in_out, N) for _ in range(runs)])
    return t, runs

def calc_res(in_outs):
    res = []
    for in_out in in_outs:
        for num_of_observers in [1, 10, 100, 1000, 10000]:
            for field in "BH":
                for obj_type in objs_params:
                    t, runs = get_time_avg(obj_type, field, in_out, num_of_observers)
                    d = {
                        "obj_type": obj_type,
                        "num_of_observers": num_of_observers,
                        "field": field,
                        "time": t,
                        "runs": runs,
                        "in_out": in_out,
                    }
                    print(f"{str(d):1000}", end="\r")
                    res.append(d)
    df = pd.DataFrame(res)
    df.to_csv(f"res_{'_'.join(str(v) for v in in_outs)}.csv", index=False)

# calc_res(in_outs=[None]) # magnetization-polarization branch

# calc_res(in_outs = ["auto", "inside", "outside"]) # inside-outside-getM-getJ

df1 = pd.read_csv("res_None.csv")
df2 = pd.read_csv("res_auto_inside_outside.csv")
df = pd.concat([df1, df2])
df["in_out"] = df["in_out"].astype(str)

px.line(
    df,
    x="num_of_observers",
    y="time",
    facet_col="obj_type",
    facet_row="field",
    color="in_out",
    line_dash="in_out",
    #log_x=True,
    #log_y=True,
    height=600
)
```

OrtnerMichael:
Hello @Alexboiboi 

I ran into some issues that are not trivially solved.

The input 'in_out' in the core functions is very troublesome because it is inconsistent and is in my opinion not needed.

`cuboid_field(field='B')` gives always the correct values, even when one mixes up outside and inside.
`cuboid_field(field='H')` gives only correct values when declaring outside and inside correctly. The inside value when one is actually outside differs by M, and similar when outside<>inside.

so far so good - a bit odd but might be consistent.

However, `cylinder_field(field='B')` gives correct values when mixing up inside<>outside for the z-component of the field, and `cylinder_field(field='H')` is doing the same for x-y components of the field.

This is ofc a result of the structure of the solution, cuboid_field computes B, while cylinder_field is an odd mixture between B and H.

**This is a huge inconsistency!** We have to declare what happens when inside is mixed up with outside.

Also, this has the potential to create wrong results in a silent way!

Do we need `inside_outside` ? This will not make any computation really faster as the bottleneck is somewhere else.

I vote for removing `inside_outside` completely as a parameter in the core fields. I see no benefit, but a lot of potential to create confusion and errors.

what do you think?

OrtnerMichael:
@Alexboiboi 

Another thing: The core functions are now split up into the actual core functions, and the functions that handle inside-outside, special input cases and so on.

The true core functions are actually the field computation functions. Would it not make more sense to have THESE at the core ?

Makes much more sense to me - this was also the original idea, that now changed by separating the core functions into multiple parts.

Alexboiboi:
> The core functions should be only the base-implementations of the explicit expressions - nothing else.

makes sense, I'll let you decide. I anyway never use them!

Alexboiboi:
> I vote for removing `inside_outside` completely as a parameter in the core fields. I see no benefit, but a lot of potential to create confusion and errors.

I agree that the benefit is very limited  for the classical field funtions. But still can greatly improve performance for `TriangularMesh` objects. We spent quite some efforts to disable/enable checks at object instantiation but we then still needed to perform the inside/outside check in `getBHMJ`. With the new `in_out` parameter on the top level it becomes possible to disable it.

OrtnerMichael (approved):
i like :)

PR #725, opened 2024-01-22
Do not trigger unit tests on docs-only changes

Alexboiboi:
There is no need to trigger unit-testing when only the docs are modifed. We already have the docs build triggered on pull requests and this hast to pass in order to merge.


OrtnerMichael (approved):
ok

PR #734, opened 2024-02-23
Replace `assert np.allclose` by `np.testing.assert_allclose`

Alexboiboi:
- needs #717 

This PR rewrites the tests to make use of the more appropriate `np.testing` module. The motivation being that the error messages are much clearer and also array shape is checked in the same step.

-  `assert np.allclose` -> `np.testing.assert_allclose`
- `assert np.all` -> `np.testing.assert_array_equal`

Note that `np.allcose` and `np.testing.assert_allclose` behave slightly differently and have different default values.

```python
np.allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)
```

```python
np.testing.assert_allclose(
    a,
    b,
    rtol=1e-07,
    atol=0,
    equal_nan=True,
    err_msg='',
    verbose=True,
)
```

Where applicable, I removed the shape equality check which is already performed by `np.testing.assert_allclose`. Where the tests where failing after the change I also applied the same 'rtol=1e-05, atol=1e-08` as the default `np.allclose`. This is due to `atol=0` for `np.testing.assert_allclose`.


OrtnerMichael:
ok alles klar!

with BHJM branch we probably have to resolve some merge conflict now

Alexboiboi:
@OrtnerMichael  ready for merge

Issue #753, opened 2024-04-08
getB squeeze=False produces false dimensions when Collections are handed in and observers have more than one dim


OrtnerMichael:
### Magpylib version

5.x (Latest)

### What happened?

@Alexboiboi Check this out - there is still a bug in getB :).

The first print produces
(1, 1, 1, 2, 3)

The second print
(1, 1, 1, 1, 2, 3)

The collection wrapping the cube should not add a dimension to getB - it is handled as a single individual source.
The crazy thing is, when observers are one-dimensional (a single point) this error does not appear :)

### Code example

```Python
import magpylib as magpy

cube = magpy.magnet.Cuboid(
    polarization=(0,0,1),
    dimension=(2,2,2),
)
coll = magpy.Collection(cube)

obs = [(0,0,0)]*2
field = cube.getB(obs, squeeze=False)
print(field.shape)

field = coll.getB(obs, squeeze=False)
print(field.shape)
```

### Additional context

this templae sux a bit ... its not practical to use - you dont get a preview :

PR #755, opened 2024-04-11
Fix Collection getB - Bug

OrtnerMichael:
# Related Issues
- #753


Alexboiboi (approved):
This only happens because with have *inputs (start inputs) in getB. But I guess the fix is also fine.

PR #756, opened 2024-04-11
Fix plotting labeling of deeply nested collections

Alexboiboi:
This fixes the wrong labeling of deeply nested collections on `show` call. Label propagation was not working beyond the first level.

```python
import magpylib as magpy
cube = magpy.magnet.Cuboid(style_label="cube")
nested = magpy.Collection(cube, style_label="nested")
root = magpy.Collection(nested, style_label="root")
magpy.show(root)
```

### Before
image

### After

image


OrtnerMichael (approved):
I confirm this fixes the problem

PR #758, opened 2024-04-15
Update reso_site_notice.md

OrtnerMichael:
note on winning a small dev grant on the webpage


Alexboiboi (changes requested):
@OrtnerMichael  I would put it like this:
In April 2024, Magpylib was awarded a "small development grant" by NumFocus.

OrtnerMichael:
ok :) - how about now ?

Alexboiboi (approved):

PR #772, opened 2024-05-16
fix plot boundary error

### LINKED ISSUE #771: Very minor oddity/bug with show function when showing single straight line current
### Magpylib version

5.x (Latest)

### What happened?

I tried to run a simple test on just a wire using the following code, but the show function fails to zoom properly.

A = 90e-3
u=1e-3
w0 = magpy.current.Polyline(current=A,vertices=((0.0025,0.0025,-u/2),(0.0025,0.0025,u/2)),position=(0,0,0))
magpy.show(w0)

The show function is zoomed out in a 1m x 1m x 1m box which shows nothing. 

If you alter the line a bit to have a tilt to one side however, eg changes position from 0.25 to 0.24 one place, the show function works again.

w0 = magpy.current.Polyline(current=A,vertices=((0.0025,0.0025,-u/2),(0.0025,0.0024,u/2)),position=(0,0,0))

Alternatively once you have two currents two different places it also works again, so it most likely is only in the single current case, so really a minor and niche bug.

Just a heads up in case it is not a known issue. Otherwise appologies for spam!

### Code example

```Python
import magpylib as magpy
A = 90e-3
u=1e-3
w0 = magpy.current.Polyline(current=A,vertices=((0.0025,0.0025,-u/2),(0.0025,0.0025,u/2)),position=(0,0,0))
magpy.show(w0)
```

### Additional context

_No response_

OrtnerMichael:
- Fixes #771

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -70,10 +70,10 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v4
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.10
       - name: Install flit
         run: pip install flit
       - name: Build package
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -38,7 +38,6 @@ jobs:
           - "3.12"
           - "3.11"
           - "3.10"
-          - "3.9"
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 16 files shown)


OrtnerMichael:
hi @Alexboiboi 

Do you know why the macos tests are all failing?

br

OrtnerMichael:
ah ... i think its the doctests ... again :(

OrtnerMichael:
its the same problem we had with numerical stability where windows and macs yielded different results at sigularities because they treat floating point somehow differently - in this case resulting in different zeros

OrtnerMichael:
@Alexboiboi 

now its ready for review :)

Alexboiboi (approved):
We got there 😄

PR #774, opened 2024-05-19
Fix tests with macos CI backend + adapt to pylint and rtd version

Alexboiboi:
### Notes
This PR addresses some failing tests with the new macos CI backend (similar to numerical stability problems) and some new pylint warnings with new package version. Also sets the minimum version python to 3.10 in github workflows and RTD.

PS: When this is merged, #772  PR will only address the original display fix ;)


OrtnerMichael (approved):
looks familiar :)

PR #780, opened 2024-05-27
Fix subplot object properties propagation

### LINKED ISSUE #775: show somehow looses the polarization when displaying an object twice
simple code

```python
import magpylib as magpy

c1 = magpy.magnet.Cuboid(
    dimension=(1,1,1),
    polarization=(1,2,3),
)
c2 = c1.copy()
```

the following code shows 2 cubes with good polarization

```python
magpy.show(
    dict(objects=c1, col=1),
    dict(objects=c2, col=2),
)
```

image

However, when handing over the same object 2 times it creates the following output

```python
magpy.show(
    dict(objects=c1, col=1),
    dict(objects=c1, col=2),
)
```
image

Alexboiboi:
- Fixes #775

### DIFF
--- a/magpylib/_src/display/traces_generic.py
+++ b/magpylib/_src/display/traces_generic.py
@@ -779,24 +779,26 @@ def get_row_col_traces(flat_objs_props, extra_backend=False, autosize=None, **kw
         else:
             traces_dict[obj] = []
             rco_obj = params.pop("row_cols")
-            for rco in rco_obj:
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
fixes the problem

OrtnerMichael:
I'm a a bit confused here - I was just about to answer that its not so important :) - but it seems the fix was quick in the end ?

Issue #716, opened 2023-12-30
Add units choice to the `show` function


Alexboiboi:
### Describe the feature you'd like

Since version 5 moves to meters instead of millimeters, the scale of the 3d plots will always be way below 1 meter. It would be nice to be able to choose the units of the coordinate system.

OrtnerMichael:
I like the idea ... what would be the default ? what would be the interface ?

Alexboiboi:
When working on some magpylib v4 to v5 transcription I also see that the default units in meters is kind of ugly in most cases when we deal with mm or cm large objects. I suggest to have the same behavior as in the default style_description where the unit prefix is chosen based on the value. We can always add this later if we miss the parameterization.
@OrtnerMichael What do you think?

PR #786, opened 2024-06-04
Add `units_length` input to the `show` function

Alexboiboi:
### Related Issues
- Implements #716 

### Notes
This adds a new `units_length` parameter to the `show` function and allows the user to choose the length units for displaying Magpylib objects per-row-and-column.

**Side effects**
- The `zoom` parameter is now also per-row-and-column.
- Ranges are individual for each plot and not overall like before.
- Pyvista plots have grids visible by default to show axes labels
- colors are iterating over subplots (before, a new cycle started for each subplot)

### Examples
```python
import magpylib as magpy

c1 = magpy.magnet.Cuboid(
    dimension=(1, 1, 1),
    polarization=(1, 2, 3),
)

with magpy.show_context(c1, backend="matplotlib") as s:
    s.show(row=1, col=1, units_length="m", zoom=3)
    s.show(row=1, col=2, units_length="dm", zoom=2)
    s.show(row=2, col=1, units_length="cm", zoom=1)
    s.show(row=2, col=2, units_length="mm", zoom=0)
```
image

### Tasks
- [x] Allow length units:
  - [x]  classic ones `['ym', 'zm', 'am', 'fm', 'pm', 'nm', 'µm', 'mm', 'm', 'km', 'Mm', 'Gm', 'Tm', 'Pm', 'Em', 'Zm', 'Ym']`
  - [x] `"dm"` and `cm` (only for manuel entry, `"auto"` will not infer these
  - [x] `None` defaults to SI (m) and does not show unit in axes labels
  - [x] "auto" will infer a suitable length unit depending on the max range value for each subplot independently
  - [x] define default value (`"m"`, `"auto"` or `None`)
- [x] Adapt ranges to units per row-col-combination.
- [x] Catch inconsistent identifiers (`zoom`, `units_length`, etc. ) as per row-col basis.
- [x] Ensure all elements are correctly scaled
  - [x] paths
  - [x] arrows
  - [x] meshes
  - [x] extra generic traces
  - [x] extra non generic traces
- [x] Pass on right unit in axes labels
  - [x] matplotlib
  - [x] plotly
  - [x] pyvista
- [x] Docs: add or modify an example to reflect the changes
- [x] Testing: add unit tests to cover the common and edge cases
- [x] Changelog: document new feature


OrtnerMichael:
hello @Alexboiboi 

I think there is a misconception of this feature - or better stated - I through of a different feature :)

**What you did**
Assume that Magpylib input is SI and change axes ticks according to the demanded unit with the purpose to adjust labels to taste. This solves e.g. problems with high `zoom`.

- Missing1: `"auto"`  setting where the unit is chosen that has at least 1 digit in front of the dot.
- Question1: Why not `style_unit_length` instead of `unit_length` ?
- Discussion point1: I would not use "dm" and "cm" ... they are terrible units, would stick to the normal ones

**What I thought we want**
Replace the displayed "(m)" with an arbitrary label, e.g. "(mm)".  This would be really practical for working with arbitrary input units. The above provides a too stiff corset that forces ppl to either not use arbitrary units, or suffer wrong labels.

Something like `style_unit_label` ? or `style_label` to make it completely arbitrary, which would be helpful in cases where the axes are not called xyz - i have such a case.

- Missing2: In any case we will need a simple way to disable the unit label

Alexboiboi:
> * Missing1: `"auto"`  setting where the unit is chosen that has at least 1 digit in front of the dot.

The way it is implemented right now makes it challenging to add the "auto" mode. The traces unit scale  needs to be known in advance to create them. In the same time you need all the traces to determine the max ranges from which you infer the unit scale. 
I'll need to think of the way to do this by avoiding to create the traces twice. It'll be quite some rework but should be feasible.


> * Question1: Why not style_unit_length instead of unit_length ?
We cannot use a style here because styles are object specific and you cannot mix different units in the same scene. It would not make any sense. This feature has to be subplot specific. 

Also the use of plural for `units` is on purpose. This allows to add more units with the kw magic syntax (e.g. `{"units_length": "mm", units_polarization": "mT"} <=> {"units":{"length":"mm", "polarization":"mT"}}`


> Discussion point1: I would not use "dm" and "cm" ... they are terrible units, would stick to the normal ones

I know theses are a bit unconventional units but can be useful to show a plot for a non technical client. I would keep these, we can remove them from the docs example if you like.


> What I thought we want
Replace the displayed "(m)" with an arbitrary label, e.g. "(mm)". This would be really practical for working with arbitrary input units. The above provides a too stiff corset that forces ppl to either not use arbitrary units, or suffer wrong labels.

I have to admit that I still have a hard time to see the benefit of arbitrary units. Especially considering the possible confusion when mixing with other objects that may have other ones in the same plot. Also considering force code coming where the use of arbitrary units is anyway not possible...
In a previous discussion I believed it was making sense to use arbitrary units only in non-object-oriented/functional approach. Maybe show me an example with a clear benefit?

>Something like style_unit_label ? or style_label to make it completely arbitrary, which would be helpful in cases where the axes are not called xyz - i have such a case.
Could you show a minimal example of this??


> Missing2: In any case we will need a simple way to disable the unit label

We can add the `units_length=None` which will disable the unit label entirely and use SI by default (dimension multiplication factor = 1)

OrtnerMichael:
**Why not style_unit_length instead of unit_length ?**

*We cannot use a style here because styles are object specific and you cannot mix different units in the same scene. It would not make any sense. This feature has to be subplot specific.*

hmm ok ... is it clear from the context that we are addressing the axis labels ?

OrtnerMichael:
**I would not use "dm" and "cm" ... they are terrible units, would stick to the normal ones**

*I know theses are a bit unconventional units but can be useful to show a plot for a non technical client. I would keep these, we can remove them from the docs example if you like.*

Im not a fan, but i will not fight tooth and nail over it

OrtnerMichael:
**Replace the displayed "(m)" with an arbitrary label, e.g. "(mm)". This would be really practical for working with arbitrary input units. The above provides a too stiff corset that forces ppl to either not use arbitrary units, or suffer wrong labels.**

*I have to admit that I still have a hard time to see the benefit of arbitrary units. Especially considering the possible confusion when mixing with other objects that may have other ones in the same plot. Also considering force code coming where the use of arbitrary units is anyway not possible...
In a previous discussion I believed it was making sense to use arbitrary units only in non-object-oriented/functional approach. Maybe show me an example with a clear benefit?*

I know you are not a fan of arbitrary units but this is shocking :). I work 90% in arbitrary units because its much faster, code is easier to read and write. Only when implementing something properly that is shared with or used by multiple users I use SI.

Which is easier to read and write ?
```
# ARBITRARY UNITS EXAMPLE
sensor = magpy.Sensor(pixel=[(-.1,0,0), (.1,0,0)])
sensor.position=np.linspace((-1,0,1), (1,0,1), 100)


# SI UNITS EXAMPLE
sensor = magpy.Sensor(pixel=[(-.0000001,0,0), (.0000001,0,0)])
sensor.position=np.linspace((-1e-6, -1e-6, 1e-6), (1e-6, 1e-6, 1e-6), 100)
```

There are advantages when making things verbose, but it is at the cost of write/readability, quick development times, good code flow overview.

-----------------------------------------

Users should easily be able to set their own axis/unit labels. This should be independent of the chosen unit, which basically just adjusts the axis tick labels. It is not a contradiction to set both, `units` and `labels`.

We have to think about a labels interface though. Probably this should be a separate PR. here some thoughts:

`labels_axis="abc"`
`labels_units="mm"`

would create "a (mm)" on the x-axis.

`labels=("A 1", "B 2", "C 3")`

would set "A 1" as label on the x-axis

`labels_units_show=False` could disable the unit labels

Alexboiboi:
> **Why not style_unit_length instead of unit_length ?**
> 
> _We cannot use a style here because styles are object specific and you cannot mix different units in the same scene. It would not make any sense. This feature has to be subplot specific._
> 
> hmm ok ... is it clear from the context that we are addressing the axis labels ?

we are not just changing the labels when using `units_length` we are also changing the values of the object coordinates. 
Since this parameter is only present in the `show` function what else would it do? We could of course also explicitly add it to the function signature and to the docstring

Alexboiboi:
> **I would not use "dm" and "cm" ... they are terrible units, would stick to the normal ones**
> 
> _I know theses are a bit unconventional units but can be useful to show a plot for a non technical client. I would keep these, we can remove them from the docs example if you like._
> 
> Im not a fan, but i will not fight tooth and nail over it

This will only apply if set explicitly, the auto mode will only infer classic units. 

If we so choose to integrate the units handler to the library it would be trivial to allow any length unit (inches, 'miles' etc.. ^^)

Alexboiboi:
> I know you are not a fan of arbitrary units but this is shocking :). I work 90% in arbitrary units because its much faster, code is easier to read and write. Only when implementing something properly that is shared with or used by multiple users I use SI.

you said it yourself, you will use SI units when sharing your work. So when you work on your own with arbitrary units, how come is it important what units are shown in the plots. In this expert mode you can just ignore it. 


> Which is easier to read and write ?

Your example can be rewritten in a quite clean way also with SI ^^
```python
# SI UNITS EXAMPLE
sensor = magpy.Sensor(pixel=np.array([(-.1,0,0), (.1,0,0)])*1e-6)
sensor.position=np.linspace(((-1,0,1), (1,0,1), 100), 100)*1e-6
```

Like I said, I don't like the possible confusion for a standard user. Add messing with the labels seems a bit like a dirty trick here.
Anyway I the user whishes to changes the labels of the plot he can easily return the figure and do it there. At least, this way it is the user's responsibility.

Btw I already implemented the `units_length=None` option, which hides the labels units on the axes and also the `units_length="auto"` which infers the units from the max range and zoom level.

OrtnerMichael:
1. adding np.array does not increase readability :), in most cases you will also need an extra import numpy (not in this case)

2. ofc you can use return_fig and then edit your own plotter, but this deafeats the whole purpose of the "expert mode" which is reducing code.

3. i dont care about the standard user or any other user - its my code that noone is supposed to read, which is 90% of my coding when working on magnetic field problems.

4. just because i dont want ppl to read my codes does not mean that i do not want them to see my figures, for which i need proper units

5. its very irritating for me to have (m) on my plots. maybe other ppl can ignore it, but i cannot :).

6. the idea of implementing SI units was not to abolish scaling and enforce SI-use but to make SI the default. This feels a lot like being forced to work with SI units.

turning units off is already great :) - i can probably live with that.

fast tuning to have my own labels would be nice, but maybe we give it some time to think about it and handle this in a separate discussion.

OrtnerMichael:
*we are not just changing the labels when using units_length we are also changing the values of the object coordinates.
Since this parameter is only present in the show function what else would it do? We could of course also explicitly add it to the function signature and to the docstring*

this i don't understand.

1. why are the object coordinates changed and not just the tick values ?

2. not in the function signature because it is part of **kwargs right ? in the docstting we should note that style and unit kwargs can be given to make users aware

OrtnerMichael:
*This will only apply if set explicitly, the auto mode will only infer classic units.*

good idea

*If we so choose to integrate the units handler to the library it would be trivial to allow any length unit (inches, 'miles' etc.. ^^)*

^^ ... when extragalactic astrophysicists start using magpylib we will also need ly :D ...  the unit handler is very cool. give it a bit more time to try it out.

Alexboiboi:
> _we are not just changing the labels when using units_length we are also changing the values of the object coordinates. Since this parameter is only present in the show function what else would it do? We could of course also explicitly add it to the function signature and to the docstring_
> 
> this i don't understand.
> 1. why are the object coordinates changed and not just the tick values ?

because when you add other objects to the plot the coordinate muss match. Plotly also shows the vertices on hover. This would require an ugly hack to work and would need to be implemented separately for each backend (not sure this is even possible with pyvista).


> 2. not in the function signature because it is part of **kwargs right ? in the docstring we should note that style and unit kwargs can be given to make users aware

This is not a problem. We already have the `zoom` parameter visible. `units_length` would use the same logic.

Alexboiboi:
> 1. adding np.array does not increase readability :), in most cases you will also need an extra import numpy (not in this case)

I guess we have different opinion on that ^^

> 2. ofc you can use return_fig and then edit your own plotter, but this defeats the whole purpose of the "expert mode" which is reducing code.
> 3. i dont care about the standard user or any other user - its my code that no one is supposed to read, which is 90% of my coding when working on magnetic field problems.

I thought caring about users and user friendliness  is what Magpylib is all about🤔 

> 4. just because i dont want ppl to read my codes does not mean that i do not want them to see my figures, for which i need proper units

What about a rescaling function to have the right units before plotting? (or a `units_lengthrescale` factor)

> 5. its very irritating for me to have (m) on my plots. maybe other ppl can ignore it, but I cannot :).

> 6. the idea of implementing SI units was not to abolish scaling and enforce SI-use but to make SI the default. This feels a lot like being forced to work with SI units.
> 
> turning units off is already great :) - i can probably live with that.

`units_length=None` and `style_description_show=False` will hide all units.

> fast tuning to have my own labels would be nice, but maybe we give it some time to think about it and handle this in a separate discussion.

Could you sum up the whishes for this PR (not the extra ones that can be discussed in another PR), so we can move on?

OrtnerMichael:
*I thought caring about users and user friendliness is what Magpylib is all about*

i'm the user in this context! user friendlyness means satisfying me in this discussion!

generally i think discussions should be more about me ;))

Alexboiboi:
> * [ ]  define default value (`"m"`, `"auto"` or `None`)

Hi @OrtnerMichael, we still need to choose the default behavior (now `"m"`), please comment on this!

OrtnerMichael:
hi @Alexboiboi,

It requires 2 long commands to remove the units from legend and plot. Can we not do this easier ?

OrtnerMichael:
I think default should be `"auto"`!
I think it should not be `None`.
Any reason it should be `"m"` ?
- default SI unit
- but lots of ppl working with mm and um will have 1e-3 and 1e-6 on their figures ... better to make use of the nice auto-feature.

Alexboiboi:
> I think default should be `"auto"`! I think it should not be `None`. Any reason it should be `"m"` ?
> 
> * default SI unit
> * but lots of ppl working with mm and um will have 1e-3 and 1e-6 on their figures ... better to make use of the nice auto-feature.

Hi @OrtnerMichael,

I the overwhelming majority of cases this is nice of course. We however have an example in the docs with the pyvista streamlines, where letting the system choose leads to unexpected results (cylinder is in mm but streamlines assume meters):
image

OrtnerMichael (approved):

Issue #766, opened 2024-04-26
add "customizable animation" to docs


OrtnerMichael:
some minimal example that can be used for copy-pasting whenever we want to customize animations to make nice figures.

```python
from plotly.subplots import make_subplots
import plotly.graph_objects as go
import numpy as np
import magpylib as magpy

fig1 = make_subplots(
    rows=1, cols=2,
    specs=[[{"type": "xy"}, {"type":"scatter3d"}]])

# setting up a system with a path
n = 10
verfahrweg = 2
xs = np.linspace(-verfahrweg/2, verfahrweg/2, n)
path = np.linspace((-verfahrweg/2,0,0), (verfahrweg/2,0,0), n)

cube = magpy.magnet.Cuboid(
    dimension=(1,1,1),
    polarization=(0,0,1),
    position=path
)
sensor = magpy.Sensor(position=(0,0,1))
coll = magpy.Collection(cube, sensor)

# compute and plot B-traces
COLORS = ['palevioletred','palegoldenrod','rebeccapurple']
B0 = coll.getB()
for col,i,lab in zip(COLORS, [0,1,2], ['Bx', 'By', 'Bz']):
    fig1.add_trace(
        go.Scatter(x=xs, y=B0[:,i], line_color=col, name=lab),
        row=1, col=1)

# plot location marker
for x0,y0 in zip(xs,B0):
    fig1.add_trace(
        go.Scatter(
            visible=False,
            x=[x0]*3,
            y=y0,
            marker_size=10,
            marker_color=COLORS,
            marker_line_color='black',
            marker_line_width=1,
            name='cart position'
            ),
        row=1,
        col=1,
        )
# set visibility of initial frame true
fig1.data[3].visible=True 

# add 3D-plot to fig (compute for every instance)
for i in range(n):
    fig0 = go.Figure()
    magpy.show(
        coll,
        backend='plotly',
        canvas=fig0,
        style_path_show=False,
        style_path_frames=[i],
        markers=[[-3]*3,[3]*3],      # keep view window fixed
        style_marker_size=0.0001,
        )
    for dat in fig0.data:
        dat.visible=False
        fig1.add_trace(dat, row=1, col=2)
m = len(fig0.data)

# set initial visibility
for i in range(m):
    fig1.data[3+n+i].visible=True

# add slider
steps1=[]
for i in range(n):
    viz = [True]*3 + [False]*n + [False]*n*m
    viz[3+i]=True
    for j in range(m):
[...]

PR #798, opened 2024-06-22
Fix pyvista plotting ranges and zoom

Alexboiboi:
### Related Issues
- #766 

This fixes the plot bounds in pyvista. This was especially visible on animations where the plot ranges were not fixed.
In pyvista there is no known way of setting the bounds manually and setting the camera position by an special algorithm is not practical at all. The solution in this case is to add a transparent scatter plot, that sets the ranges correctly.

This correction also enables the zoom feature which was absent until now!

### Ranges
```python
import magpylib as magpy
import numpy as np

cylseg = magpy.magnet.CylinderSegment(
    dimension=(3, 4, 1, 0, 45),
    polarization=(0, 0, 1),
)

angles = np.linspace(0, 360, 10, endpoint=False)
cylseg.rotate_from_angax(angles, axis="z")

magpy.show(
    {"objects": cylseg, "col": 1},
    backend="pyvista",
    style_legend_show=False,
    animation=True,
    animation_duration=2,
    animation_output="test.gif",
)
```

Before |  After
:------:|:------:
test | test

### Zoom levels
```python
import magpylib as magpy
import numpy as np

cylseg = magpy.magnet.CylinderSegment(
    dimension=(3, 4, 1, 0, 45),
    polarization=(0, 0, 1),
)

magpy.show(
    *[{"objects": cylseg, "col": i+1,  "zoom":i} for i in range(4)],
    backend="pyvista",
    style_legend_show=False,
)
```

Before (also a camera issue) |  After
:------:|:------:
image | image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased] - YYYY-MM-DD
 - Added `units_length` input to the `show` function to allow displaying axes with different length units. This parameter can be set individually for each subplot. ([#786](https://github.com/magpylib/magpylib/pull/786))
 - Small documentation and Readme improvement. Change naming from "explicit expression" to "analytical expression" as described in ([#794](https://github.com/magpylib/magpylib/issues/794)).
+- Fix Pvyvista plot bounds on animation. Also enables `zoom` feature which was not working until now. ([#798](https://github.com/magpylib/magpylib/pull/798))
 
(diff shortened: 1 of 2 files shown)


Alexboiboi:
@OrtnerMichael ready to be merged

OrtnerMichael (approved):
works

PR #799, opened 2024-06-23
Fix canvas layout being modified even when user-provided

Alexboiboi:
This falls in line with the idea that when a user provides a canvas, Magpylib writes the data into it but should not update the layout.

- [x] The new `canvas_update` (default=`None`) parameter is added to the `show_function` 
  - [x] `None` updates layout only if canvas is not user-provided
  - [x] `True` -> always updates
  - [x] `False` -> never updates

> [!NOTE]
>  plotly: for animation, the slider will still be shown in all cases

### Matplotlib

```python
import magpylib as magpy
import numpy as np
import matplotlib.pyplot as plt

loop = magpy.current.Circle(current=1, diameter=0.1)
sens = magpy.Sensor(position=np.linspace((0, 0, -0.4), (0, 0, 0.4), 100))

fig = plt.figure(figsize=(10, 4))
ax = fig.add_subplot(111, projection="3d")

magpy.show(loop, sens, canvas=ax)
magpy.show(loop, sens, backend="matplotlib")
```

User Canvas  |  Auto Canvas
:------:|:------:
image | image

### Plotly
```python
import magpylib as magpy
import numpy as np

loop = magpy.current.Circle(current=1, diameter=0.1)
sens = magpy.Sensor(position=np.linspace((0, 0, -0.4), (0, 0, 0.4), 100))

fig = go.Figure()

magpy.show(loop, sens, canvas=fig)
fig.show()
magpy.show(loop, sens, backend="plotly")
```
User Canvas  |  Auto Canvas
:------:|:------:
image | image

### Pyvista
```python
import magpylib as magpy
import numpy as np
import pyvista as pv

loop = magpy.current.Circle(current=1, diameter=0.1)
sens = magpy.Sensor(position=np.linspace((0, 0, -0.4), (0, 0, 0.4), 100))

plotter = pv.Plotter()
plotter.show_axes = False

magpy.show(loop, sens, canvas=plotter)
plotter.show()
magpy.show(loop, sens, backend="pyvista")
```
User Canvas  |  Auto Canvas
:------:|:------:
image | image


OrtnerMichael:
Hi Alex,

I love this !

#plotly slider ... not sure I understand what this implies ?

#canvas_update
default should be `"auto"` or `None` ?
   - `None` should hand over formatting except when own canvas is provided. 
   - `True` should also update own canvas (like it is now on the main branch). 
   - `False` should ignore all formatings for all show outputs.
This makes a lot of sense to me.

Alexboiboi:
> plotly slider ... not sure I understand what this implies ?

If you plot an animation into an existing canvas, it would result in no control over it to be shown.

Alexboiboi:
do we really need tests for this feature?

Alexboiboi:
what about the docs?

OrtnerMichael:
Hi @Alexboiboi 

With `canvas_update=False` Matplotlib and Pyvista still show the legend - which is not default behavior I think.

OrtnerMichael:
As discussed ... 

`fig = magpy.show(loop, sens, return_fig=True, canvas_update=None)`

updates the canvas with `return_fig` with kwarg `canvas_update=None`. This I find a bit strange. Would it not make more sense to have `canvas_update="auto"` by default ? Then the user knows that this is some behavior we turn on and of as we think is best. `True`, `False`, `"auto"` we use regularly no ?

OrtnerMichael:
**There is a bug**

`magpy.show(loop, sens, backend="matplotlib")`

does not generate a figure. The `plt.show()` must be missing. It works fine for the other two backends.

OrtnerMichael:
concerning the slider: I think the way it is now is ok.

Alexboiboi:
> `fig = magpy.show(loop, sens, return_fig=True, canvas_update=None)`
> 
> updates the canvas with `return_fig` with kwarg `canvas_update=None`. This I find a bit strange. Would it not make more sense to have `canvas_update="auto"` by default ? Then the user knows that this is some behavior we turn on and of as we think is best. `True`, `False`, `"auto"` we use regularly no ?

The `return_fig` should only tell if the figure object is returned. There is no side influence.

In agree that we can make the `canvas_update` default value be "auto" if it is less confusing

Alexboiboi:
> **There is a bug**
> 
> `magpy.show(loop, sens, backend="matplotlib")`
> 
> does not generate a figure. The `plt.show()` must be missing. It works fine for the other two backends.

fixed

Alexboiboi:
> With `canvas_update=False` Matplotlib and Pyvista still show the legend - which is not default behavior I think.

This is a bit tricky. The showlegend is a per trace parameter (comes from plotly) and not a legend one. But as far I understand matplotlib functions a bit similarly but you can deactivate the legend. Since we have the style_legend_show parameter, should we keep the current behavior?

OrtnerMichael (commented):
image

why do you import magpylib before numpy ?

OrtnerMichael:
>> With canvas_update=False Matplotlib and Pyvista still show the legend - which is not default behavior I think.

> This is a bit tricky. The showlegend is a per trace parameter (comes from plotly) and not a legend one. But as far I understand matplotlib functions a bit similarly but you can deactivate the legend. Since we have the style_legend_show parameter, should we keep the current behavior?

I think we keepit as it is now

OrtnerMichael (approved):
looks good on my side.

Please address the choice of input order before merging.

Alexboiboi:
> image
> 
> why do you import magpylib before numpy ?

When I edited the docs code in jupyterlab I ran the jupyterlab-code-formatter extension which runs black and isort (like in our library). So this follows the isort rules ([isort](https://pycqa.github.io/isort/))

PR #820, opened 2024-10-30
Force and issue template

OrtnerMichael:
- remove issue templates that noone uses
- add docs for force package v0.3.0 and v0.3.1
- I know this should be two separate issues alex


fslanovc (approved):
nice documentation

PR #824, opened 2025-01-04
Optimize celv function in special_cel.py

p-sira:
# Related Issues
Remove unneccesary copy in celv function since kk is already copied from k in lines above.


OrtnerMichael (approved):
Correct - this operation is not necessary as a the multiplication of two numpy views always creates a copy. Thank you for pointing this you. You will probably find a few more of such imperfections throughout our code :).

OrtnerMichael:
oje - the doctests are failing ... i'll fix those up in a seperate branch before merging this one

PR #825, opened 2025-01-06
Fix doctest rounding mismatches

OrtnerMichael:
Display of all digits has often led to issues with the doctests.
Now we get rid of this using `with np.printoptions(precision=3)`

Also reducing examples where there are too many and adding some other examples where there are none.

Also editing a few docstrings on the way

Also removing in_out from Collection

Also fixing some bugs where core function would not return the proper fields (some factors missing)


OrtnerMichael:
@Alexboiboi 

Finally its working !
Please review and merge !

magpylib/_src/obj_classes/class_Collection.py
```diff
@@ -804,26 +773,27 @@ def getJ(
             `numpy.ndarray` object is returned. If 'dataframe' is chosen, a `pandas.DataFrame`
             object is returned (the Pandas library must be installed).
```
Alexboiboi, line 775: Why removing this in this PR ??

tests/test_field_cylinder.py
```diff
@@ -204,9 +204,7 @@ def test_cylinder_tile_slanovc(inputs, H_expected):
         "observers": inputs["obs_pos"],
         "dimensions": inputs["dim"],
     }
-    H = (
-        magnet_cylinder_segment_Hfield(**inputs_mod) / 4 / np.pi * 1e7
-    )  # factors come from B <->H change
+    H = magnet_cylinder_segment_Hfield(**inputs_mod)  # factors come from B <->H change
```
Alexboiboi, line 207: How was this not failing before ??

Alexboiboi (approved):
Rounding fixes are ok, thanks for the effort!

see remarks for the other changes.

I'll approve this, up to you if you want to make some more changes!

magpylib/_src/obj_classes/class_Collection.py
OrtnerMichael: This was not removed but moved up. There was a docsting bug such that "output" was below "Returns".

tests/test_field_cylinder.py
OrtnerMichael: I modified the core functions of cylinder segment because it returned H without this factor.

PR #833, opened 2025-03-17
Current sheet full

fslanovc:
# Related Issues


Alexboiboi:
@fslanovc do you need help to resolve merge conflicts?

OrtnerMichael:
this needs thorough review in terms of functionality and stability

fslanovc:
Many thanks for the review!
Concerning the vertices, This was actually a bug and should have worked already. Now it should work and the return value is always (0,0,0) within the whole sheet itself. This is now also written in the docstring.
Also found another minor bug with the indents.
Now everything should be working and from my side it's ready for the release!

OrtnerMichael:
@fslanovc  When I started implementing the object oriented interface I noticed that for practical reasons in the TriangleStrip case the current must always flow in parallel to one of the sides.

Would this simplify the computation a lot ? Is it worth it to implement this as a special case ?

No matter what we will also have a current surface object where all orientations will be needed.

fslanovc:
@OrtnerMichael 
Yes I believe TriangleStrip() should be fine...at least it is similar to TriangularMesh().
This will indeed simplify the calculation, because in all the formulas of the elementar current sheet, one can assume jv = 0. Strangely I'm not even sure if this will lead to any significant performance boost, because this variable appears only in some simple sums. We can try it out at some point, but I would give it rather low priority.
Screenshot 2025-07-02 101844

fslanovc:
@OrtnerMichael @Alexboiboi I know tested `TriangleStrip()` and `TriangleSheet()` - looks very good, useful and also great and interesting test cases (e.g. vs. cuboid magnet)! Really a nice new feature - great work also combining all this with the object oriented interface!! :)

OrtnerMichael:
hi @fslanovc and @Alexboiboi 

I took a lot of time adding 5 examples to the docs (see above).
They are ready now - here is a link to the built docs.

https://magpylib.readthedocs.io/en/current_sheet_full/

Would be great if you have time to review them briefly.

Two things are now missing
1. The graphical representation of the current flow
2. Redoing some example plots once pixel_field is ready

fslanovc:
[example method of images](https://github.com/magpylib/magpylib/pull/833/commits/f7a3d5e1451a1a7857bd0375cdde224814f2315c)
Great - I believe we should also make use of all the nice images to write a short paper on the mirror effect, the application of the holding force and a hint to the magpy.force package.

fslanovc:
[add planar conductors example](https://github.com/magpylib/magpylib/pull/833/commits/e510c8172655585fa759a095ed4f76a32831327d)
Looks also good.
From my point of view, all new examples look good and can be released!

PR #843, opened 2025-05-06
Move optional-dependencies to dependency groups

purepani:
Optional dependencies are user facing while dev groups are meant to be for developers.
Additionally, the tool.uv.dev-dependencies option is now a legacy option, per https://docs.astral.sh/uv/concepts/projects/dependencies/#legacy-dev-dependencies


Alexboiboi:
Hi @purepani,
I think this goes in the right direction, thanks.  One thing though is that only a recent version of pip (25.1) supports dependency groups, but since we are targeting developers, I guess this is ok.

Just added a few more changes to the PR to adapt changes in the code base.

Issue #792, opened 2024-06-12
Support Array API Standard for Python.


purepani:
Numpy 2.0 is about to come out in 4 days, so making sure support for that would be good. It would also be good to try and have support for the [Array API standard](https://data-apis.org/array-api/latest/) so that more array backends could be used as needed(which would allow for things like jax to be used), though this might be a slightly longer term goal.

Alexboiboi:
Hi @purepani,

Thank you for bringing that topic up!

The Numpy 2.0 support is addressed in #795 and a release will follow promptly.

I don't think we will work on a Array API support any time soon, so we will keep this issue open in the mean time.

purepani:
I will also link the tracking issue from scipy here, since it's probably needed first given that it's a dependency
https://github.com/scipy/scipy/issues/18867

OrtnerMichael:
Hey @purepani,

I forgot to answer - using the array API standard is a great suggestion! If I understand correctly it would allow a user to select e.g. Numpy or PyTorch or several other array frameworks for the computation.

This would immediately solve one of our main issues - that we cannot do cuda, which would be really helpful for all discretization issues that are required for magplib-material-response and magpylib-force, optimization problems and large scale parameter sweeps.

lucascolley:
> If I understand correctly it would allow a user to select e.g. Numpy or PyTorch or several other array frameworks for the computation.

Yep! The goal with SciPy is to have all computation be done with the library of the array input where possible, and to fall back to NumPy for intermediate calculations otherwise.

The extent of CUDA support will depend on which parts of SciPy you use - anything written in pure Python should become portable across array libraries quite easily, but parts which use compiled code are quite tricky, especially if there are not alternative implementations available for the other libraries.

purepani:
Hey @Alexboiboi and @OrtnerMichael  
Do you know how feasible it would be to manually implement the algorithms through the array api here? I haven't looked through the library yet to see how feasible that would be. 
I imagine the main problem would be that you use the scipy `Rotation` class as part of the API interface, but I think it would be fairly easy to just wrap that object and implement any needed operations through only the array API.
I believe there's also a [PR](https://github.com/scipy/scipy) for the the spatial module in scipy that we could borrow some code from for a wrapper if needed(or just rewrite since rotation is generally pretty simple).

(I'm mostly asking in case I want to try making something like this).

[...]

Issue #732, opened 2024-02-22
core function tests based on FE


OrtnerMichael:
add FE-based core tests

PR #844, opened 2025-05-06
Array API support for core functions

purepani:
This PR will track the status of array api support per https://github.com/magpylib/magpylib/issues/792.
TODO:
- [ ] Fields
  - [x] Dipole
  - [x] Sphere
  - [x] Cuboid
  - [x] Circle
  - [x] Cylinder
  - [x] Cylinder Segment
    - [ ] Check docstrings for changes that were made unintentionally for find and replace `np.` with `xp.` and `len(r)` with `r.shape[0]`, and maybe other things that I forgot about finding and replacing.  
  - [x] Polyline
  - [x] Tetrahedron
  - [x] Triangle
  - [x] Triangular Mesh
- [x] Special Functions
  - [x] Cel
  - [x] El3
  - [x] Elliptical E (from scipy)
  - [x] Elliptical K (from scipy)
 - [ ] `getB`/`getH`/`getM`/`getJ`
 - [ ] Fix use of `xp.linalg` functions to be guarded under an if statement, or using a utility. Currently, the arrays are assumed to have the linalg extension, but this is not guaranteed by the api.
 - [ ] Implement a better conversion to float that isn't cast to `xp.float64` for everything.
 - [ ] Implement a `lazy_while` for the elliptic integrals.

So far, I've implemented support for the dipole core function, and also modified the test to use `array-api-strict` instead of `numpy`.

I had to change the core function to use masks to avoid division by 0 instead of ignoring warnings, since otherwise there is no way to ignore the warnings in a general way.  I haven't tested this yet with any API backends.

I'm planning on using this PR to collect all the commits for the core functions, but I am also able to split them into separate PR, since that will be easier to review(hence the empty dev commit). As of right now, the dipole could be merged on it's on after review, since it doesn't change the behavior of the library(though an argument is added to the dipole function for the array backend). I can create a separate PR for each function as they are implemented if wanted, but I'll keep this draft PR to track it's overall progress.

### DIFF
--- a/src/magpylib/_src/fields/field_BH_cuboid.py
+++ b/src/magpylib/_src/fields/field_BH_cuboid.py
@@ -6,9 +6,11 @@ magnetized Cuboids. Computation details in function docstrings.
 from __future__ import annotations
 
 import numpy as np
+from array_api_compat import array_namespace
 from scipy.constants import mu_0 as MU0
 
 from magpylib._src.input_checks import check_field_input
+from magpylib._src.array_api_utils import xp_promote
 
 # pylint: disable=too-many-statements
 
@@ -79,9 +81,14 @@ def magnet_cuboid_Bfield(
 
     Cichon: IEEE Sensors Journal, vol. 19, no. 7, April 1, 2019, p.2509
     """
-    pol_x, pol_y, pol_z = polarizations.T
-    a, b, c = dimensions.T / 2
-    x, y, z = np.copy(observers).T
+    xp = array_namespace(observers, dimensions, polarizations)
+    observers, dimensions, polarizations = xp_promote(
+        observers, dimensions, polarizations, force_floating=True, xp=xp
+    )
+
+    pol_x, pol_y, pol_z = (polarizations[:, i] for i in range(3))
+    a, b, c = ((dimensions / 2)[:, i] for i in range(3))
+    x, y, z = (observers[:, i] for i in range(3))
 
     # avoid indeterminate forms by evaluating in bottQ4 only --------
     # basic masks
@@ -95,10 +102,10 @@ def magnet_cuboid_Bfield(
     z[maskz] = z[maskz] * -1
 
     # create sign flips for position changes
-    qsigns = np.ones((len(pol_x), 3, 3))
-    qs_flipx = np.array([[1, -1, -1], [-1, 1, 1], [-1, 1, 1]])
-    qs_flipy = np.array([[1, -1, 1], [-1, 1, -1], [1, -1, 1]])
-    qs_flipz = np.array([[1, 1, -1], [1, 1, -1], [-1, -1, 1]])
+    qsigns = xp.ones((pol_x.shape[0], 3, 3), dtype=xp.float32)
+    qs_flipx = xp.asarray([[1, -1, -1], [-1, 1, 1], [-1, 1, 1]], dtype=xp.float32)
+    qs_flipy = xp.asarray([[1, -1, 1], [-1, 1, -1], [1, -1, 1]], dtype=xp.float32)
+    qs_flipz = xp.asarray([[1, 1, -1], [1, 1, -1], [-1, -1, 1]], dtype=xp.float32)
     # signs flips can be applied subsequently
     qsigns[maskx] = qsigns[maskx] * qs_flipx
     qsigns[masky] = qsigns[masky] * qs_flipy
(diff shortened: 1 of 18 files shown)


purepani:
I just implemented the sphere. I implemented the inside/outside with masks. While, as noted in the comments, this might not be ideal for CPU(I'm not quite sure about this, but I assume you investigated that and found it to be true), It will make a good difference for the gpu. Ideally, these core functions should be implemented fully with array operations, so that backends like `jax` can compile them super efficiently.
I will add a checklist at the top for the different functions.

purepani:
Also note that each commit is individually reviewable, and I plan to keep it that way :)

purepani:
Additionally, if someone is willing to work on https://github.com/magpylib/magpylib/issues/732, I would feel a lot more confident in my translations here.

purepani:
Added cuboid, polyline, and triangle.
I'm fairly certain these aren't going the most efficient way to implement these, but that can be improved later.

Alexboiboi:
Hi @purepani, 

Thank you very much for starting this ! 
I'm currently not able to spend much time reviewing this but I soon will. 

Not sure yet if we should split this in multiple PRs but in the first step, modifications should only apply to the core functions ! Later when SciPy functions (Rotation) is implemented in an array-api compatible way we can start working the the general functional interface. And finally see if this can be applied all the way through the object oriented interface.

BR, Alex-

purepani:
I have the commits organized so that it's super easy to create multiple PRs if needed, so I can make them super easily if you'd prefer that! Or I can leave it all here! Just let me know if you want one or the other.

OrtnerMichael:
Hey @purepani,
Its the first time someone outside of the core team (me and alex ;D) is attempting a major thing like this - so thanks a lot for starting this. I will take time to review and help where I can.

Concerning the FE checks #732: Not really relevant. Our implementations are all properly tested - the FE tests were simply thought as a "second guarantee" that things are ok. In any case FE is massively lacking the precision the analytical method offers.

I think @Alexboiboi is right that we should only start on the core functions for now, top level `getB` should have a parameter `numerical_backend='numpy'` or something like this that is by default set tom `'numpy'` and allows one to select respective other packages.

Concerning special functions: in some packages some are well implemented and ready in compiled form - in which case we should make use of them. In other cases we should have our own (which will never be as fast). Not 100% sure how to handle this ?

br michael

purepani:
> I think @Alexboiboi is right that we should only start on the core functions for now, top level getB should have a parameter numerical_backend='numpy' or something like this that is by default set tom 'numpy' and allows one to select respective other packages.
Just to respond to this, the current way the array api works is that any array `x` from an array has a function `xp = x.__array_namespace__()`, such that `xp` acts like the array "module,"(i.e. in a similar way to numpy). So such a parameter for the numerical backend isn't necessary.

Some of the things that we should look to the current efforts in scipy are listed here: https://data-apis.org/array-api/latest/design_topics/index.html
I particularly want to figure out how to deal with lazy vs eager execution. Right now I'm optimizing for lazy execution(like in libraries like `jax`) but this will cause performance degradation in eager execution libraries(like numpy).

src/magpylib/_src/fields/field_BH_cuboid.py
```diff
@@ -46,11 +47,11 @@ def magnet_cuboid_Bfield(
     >>> import numpy as np
     >>> import magpylib as magpy
     >>> B = magpy.core.magnet_cuboid_Bfield(
-    ...    observers=np.array([(1,1,1), (2,2,2)]),
-    ...    dimensions=np.array([(1,1,1), (1,2,3)]),
-    ...    polarizations=np.array([(0,0,1), (.5,.5,0)]),
+    ...    observers=xp.array([(1,1,1), (2,2,2)]),
+    ...    dimensions=xp.array([(1,1,1), (1,2,3)]),
+    ...    polarizations=xp.array([(0,0,1), (.5,.5,0)]),
```
lucascolley, line 52: `xp.array` does not exist in the standard, you probably want `xp.asarray`

src/magpylib/_src/fields/field_BH_cuboid.py
```diff
@@ -46,11 +47,11 @@ def magnet_cuboid_Bfield(
...
     ... )
-    >>> with np.printoptions(precision=3):
+    >>> with xp.printoptions(precision=3):
```
lucascolley, line 54: this is not in the standard

src/magpylib/_src/fields/field_BH_dipole.py
```diff
@@ -53,43 +55,40 @@ def dipole_Hfield(
...
-    # when r=0 return np.inf in all non-zero moments directions
-    mask1 = r == 0
-    if np.any(mask1):
-        with np.errstate(divide="ignore", invalid="ignore"):
-            H[mask1] = moments[mask1] / 0.0
-            np.nan_to_num(H, copy=False, posinf=np.inf, neginf=-np.inf)
+    xp = array_namespace(observers, moments)
+    r = xp.sqrt(xp.vecdot(observers, observers))[
+        :, xp.newaxis
+    ]  # faster than np.linalg.norm
```
lucascolley, line 61: x-ref https://github.com/scipy/scipy/blob/9b6dff3fd5aa95b557ff74d02a92e946c5f9378e/scipy/_lib/_array_api.py#L459-L482

src/magpylib/_src/fields/field_BH_cuboid.py
purepani: Huh, I thought I removed these

src/magpylib/_src/fields/field_BH_cuboid.py
purepani: Same here, I thought I removed them

lucascolley:
@Alexboiboi @OrtnerMichael @purepani of course feel free to proceed however you would like! But I would suggest the following integration strategy:

- keep one branch with all changes that seem to be required (e.g. this one) so that we can see the total extent of changes
- keep changes to separate modules in separate commits so they can be broken out into PRs at a later stage
- if any changes make sense to do even for just NumPy (e.g. refactoring), merge them first in separate PRs to reduce the diffs
- if it seems like you need to write any helpers which would be of general use rather than specific to magpylib, open an issue at array-api-extra. It may be that a solution already exists in SciPy/scikit-learn, or that something you write here can be upstreamed to array-api-extra
- once everything starts to look relatively clean and feasible, have a think about how you want to test, and look into following what we have upstream
- at that stage, think about the release strategy. Do you want to release a new major version with array-api-extra and array-api-compat as required dependencies? Or would you rather vendor them and gradually expand support in an experimental mode first?
- once all of this is done, start breaking out PRs from the umbrella branch for individual review and merge

lucascolley:
> > I think @Alexboiboi is right that we should only start on the core functions for now, top level getB should have a parameter numerical_backend='numpy' or something like this that is by default set tom 'numpy' and allows one to select respective other packages.
> Just to respond to this, the current way the array api works is that any array x from an array has a function xp = x.__array_namespace__(), such that xp acts like the array "module,"(i.e. in a similar way to np). So such a parameter for the numerical backend isn't necessary.

Yes, in a catchphrase the pattern this work enables is "array type in === array type out". I highly recommend reading https://github.com/scipy/scipy/issues/18286 to understand the design we are implementing in SciPy.

purepani:
I disabled the "Allow edits by maintainers" for now since I keep forgetting to run the precommit hook manually at the moment(I use [jj](https://jj-vcs.github.io/jj/latest/) which doesn't support git precommit hooks), and the automatic commit from the bot is a bit annoying. Obviously I'll try and format stuff when I remember to run ruff, but it's a bit annoying to deal with the bot adding commits every time I push.

purepani:
For getting elliptic functions working, I used `array_api_strict.lazy_apply` in order to implicitly convert to numpy arrays for the scipy functions. This means that for backends that run on the gpu, this won't work without copy.

The reason I chose to do this right now instead of writing a python implementation for supporting gpus was because 
A. It'll be way slower for CPU code
B. It wouldn't work with lazy backends like `jax` since the implementation uses a value dependent while loop(to test for tolerance).

The solution in jax would be to use `jax.lax.while_loop`, but obviously we can't directly use that here, so a utility function `lazy_while` that dispatches to `jax.lax.while_loop` for jax. Something similar will probably need to be done for [`dask`](https://github.com/dask/dask), but I have never used it before, so I have no clue how an implementation for it would work.

purepani:
As painful as doing the cylinder segment was(why are there so many special cases :sob:) it that is now complete. I did a lot of basic find and replace for `np. -> xp.` and `len(r) -> r.shape[0]`, so there may be some casualties in parts of the documentation that should probably be left as it was originally, but I really wanted to push this after the test passed so I'll fix that later.

I should note that for the polyline source, the test for `current_vertices_field` currently fails, due to the array api lacking an `xp.split` function. I don't think this is something that should be relied on for this function, since it's a dynamic indexing, and lazy backends probably won't support that, but I don't have a great solution at the moment for alternatives. 

I haven't thought so hard about this yet, and don't have a suggested solution in mind, mostly because that cylinder segment tired me out, but I'll try and come up with suggestions, and feel free to give your own. I'm guessing the best way to do this will be a breaking API change to that function, but there may be other options.

purepani:
Also another note/question to all of you.
Numpy does a lot of automatic upcasting from int to float which the array api requires to be manually done. Right now I've just been brute forcing casting to float64 without any checks if I ran into an error. How should I be casting, and how aggressively should I be doing so?

In the spirit of array_in=array out, I think that array_out should take on the maximum byte size float dtype that is in the argument(so if `observers.__dtype__ = xp.float32` and `positions.__dtype__=xp.int32`, then the out array should be of dtype `xp.float32`, but how should we handle the case when none of them are floats? Additionally, how should we handle doing those internal conversions to floats, since the conversion can cause precision loss during computation.

purepani:
> I should note that for the polyline source, the test for current_vertices_field currently fails, due to the array api lacking an xp.split function. I don't think this is something that should be relied on for this function, since it's a dynamic indexing, and lazy backends probably won't support that, but I don't have a great solution at the moment for alternatives.

I just realized this isn't part of the public api, so breaking the public api doesn't matter :smile:

lucascolley:
RE: promotion, check out https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L553-L558

lucascolley:
there is a "crude" version of `split` at https://github.com/scipy/scipy/pull/22122#discussion_r1916728728. It was suggested for xpx at https://github.com/data-apis/array-api-extra/issues/97, but no discussion so far.

purepani:
@Alexboiboi @OrtnerMichael 
I realized I went beyond the scope of "just the core functions," since I erroneously assumed that everything in `src/magpylib/_src/fields` was part of the core functions. Should I make a PR for all the core functions specifically?

purepani:
@lucascolley for that promotion function, do you suggest just using it on everything to promote everything to floats right at the beginning? Or do you suggest using it only just before any required float support for specific functions(i.e. `xp.sqrt`)

purepani:
Makes sense! I doubt there's any significant performance difference anyway, and if there is, that's certainly under the "premature optimization" trap.

OrtnerMichael:
hi @purepani 

Sorry for the missing feedback from our side. So far I was not able to review all your PR's. I'm about to complete a publication and will then have more time to invest...probably starting next week.

As this touches the computational core I'll give it a very very thorough review. We should go through this function by function. which core function / PR do you suggest we start with ?

lucascolley:
@OrtnerMichael I wrote a suggested review strategy here: https://github.com/magpylib/magpylib/pull/844#issuecomment-2868709593

purepani:
Hey @lucascolley 
Dask doesn't support the full linalg extension for the array api, but it also seems like they implement part of the linalg extension, so checking for the linalg extension with `hasattr(xp, 'linalg')` doesn't work. I just wanted to make sure that my understanding here is correct before making an issue in dask.

lucascolley:
@lithomas1 do you now if there is an issue or some other status somewhere for `dask.linalg` in array-api-compat?

lucascolley:
> Per https://data-apis.org/array-api/latest/extensions/index.html, I believe the module should probably not exist under that name.

In an ideal world, yes. Unfortunately that guidance can't really hold for existing array libraries. I guess we could look at some alternative for array-api-compat, but not sure what that would look like.

purepani:
Maybe a function in `array_api_compat` which correctly gives the extension support? Or maybe such a function should be part of the spec?

lucascolley:
There is https://github.com/data-apis/array-api-compat/blob/main/array_api_compat/dask/array/linalg.py, but IIUC it is a WIP and somewhat limited by what is implemented in Dask proper.

Alexboiboi:
Closing this as the umbrella tracker — its job has moved to #792, whose body now carries the checklist, the contract decisions that have to be settled before parallel kernel ports resume, the eight-step sequence, and the disposition of each 2025 branch.

**This is not a rejection of the work.** The discussion here is the most valuable single thread on the topic and stays linked from #792 — in particular @lucascolley's integration strategy (#844 (comment)), the dtype-promotion thread, the `linalg` extension question, and the lazy-versus-eager exchange. The sequence in #792 is built from it.

A PR is a poor tracker: it conflicts, it shows a 2894-line diff nobody intends to merge, and GitHub counts it as pending work. An issue does not.

@purepani — see the comment on #792 for where things go from here, and for three answers we owed you from May 2025.

PR #859, opened 2025-05-20
Adds core tests

purepani:
Adds tests for the core functions.
I just evaluated the functions given as examples in the docstring, and use those for tests.
I will switch this to use `array-api-strict` instead of numpy for the array api, but I figured it's better to just have this in general.


purepani:
Not sure why this closed

purepani:
Not sure why this closed

PR #860, opened 2025-05-20
Adds core tests

purepani:
Adds tests for the core functions.
I just evaluated the functions given as examples in the docstring, and use those for tests.
I will switch this to use array-api-strict instead of numpy for the array api, but I figured it's better to just have this in general, so I'm adding this as a separate PR

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -3,7 +3,17 @@ from __future__ import annotations
 
 import numpy as np
 
-from magpylib._src.fields.field_BH_sphere import magnet_sphere_Bfield
+from magpylib.core import (
+    current_circle_Hfield,
+    current_polyline_Hfield,
+    dipole_Hfield,
+    magnet_cuboid_Bfield,
+    magnet_cylinder_axial_Bfield,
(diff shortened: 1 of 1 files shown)


purepani:
@Alexboiboi @OrtnerMichael 
This is able to be merged as is since this just adds some tests for the core functions. They don't cover every case for every core function, but it's at least a start, and makes it a bit more convenient to test the array api rewrite.

purepani:
Additionally, these are the tests for the exact functions I'm converting, whereas all the other tests so far(at least as far as I can tell) are further up in the chain of abstraction, so I would only be able to test them with numpy, unless I also converted more of the code. Having these tests lets  me test against the minimal amount of code necessary to convert the lowest level to the array API standard

purepani:
@Alexboiboi would it be possible for y'all to vectorize the cylinder segment computation as mentioned in the comments? I don't quite have the time to understand how everything fits together in that file for me to do it super quickly, though I can eventually do it. 

The cylinder segment is the only one left that needs to be fitted to work for lazy backends.

PR #869, opened 2025-07-08
Fix typo in axis label of example.

wittymap:


OrtnerMichael (approved):
correct - thank you for fixing this

PR #871, opened 2025-07-14
Fix building docs

Alexboiboi:
This PR fixes building docs with the `nox -s docs` command. It now includes the `docs` dependency group to the `dev` group.

Also removes `from __future__ import annotations` since it is not needed anymore for python 3.11+

TODOs:
- [x] include `docs` group  in `dev` dependency-group in `pyproject.toml`
- [x] fix typing where needed
- [x] fix plotly plots that are not shown in the docs

### DIFF
--- a/tests/test_input_checks.py
+++ b/tests/test_input_checks.py
@@ -718,8 +718,8 @@ def test_input_rotate_axis_bad(axis):
 @pytest.mark.parametrize(
     "observers",
     [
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
looks good to me - docs now building without problems (except for all those warnings when we build ...  but i'll take care of those evil blank lines and typing issues in the docstrings)

PR #884, opened 2025-08-08
100% Success on running unit tests, added 2d Ellipsoids

analog-wizard:
Added (as mentioned in the title), the ability to draw 2d ellipsoids, and 'expanded' the framework to allow for the calculation of their magnetic field. The underlying mathematical model allows for the representation for circles with multiple radii, so I expanded the present definition of 'current.Circle' to include the ability to do so. This functionality may be better suited for a new, separate object named (to the effect of) 'current.Ellipse', where there may be some hierarchical relation between the two.

This modification means that 'current.Circle' may now take array_types (and now negatives as bonus) as a 'diameter' parameter

Unfortunately, (and against my best efforts), Nox was unable to run due to a dependency conflict with the present development package (version "0.1.dev1901+gb7e0628") and 'magpylib-material-response' (required version ">=5.0"). I am not familiar with Nox, and despite googling, I have found no solution (as of yet). I have, however, manually run the pytests and have (as mentioned) a 100% pass rate. I have attempted to maintain the "spirit" of the repository in naming conventions, style, and structure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,8 @@ dependencies = [
     "scipy>=1.8",
     "matplotlib>=3.6",
     "plotly>=5.16",
+    "anywidget",
+    "pandas"
 ]
 
 [dependency-groups]
--- a/src/magpylib/_src/display/traces_core.py
+++ b/src/magpylib/_src/display/traces_core.py
@@ -6,6 +6,7 @@
 # pylint: disable=too-many-nested-blocks
 # pylint: disable=cyclic-import
 
+import numbers
 import warnings
 from itertools import combinations, cycle
 from typing import Any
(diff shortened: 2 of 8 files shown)


OrtnerMichael:
dear @analog-wizard,
Thank you for this contribution and sorry for not getting back to you faster. I will review it in the next days. As you mention, adding a new Class `Ellipsis` makes probably more sense - but give me a few more days to provide a proper review.
br michael

OrtnerMichael (commented):
Hey @analog-wizard 

As far as I understand, you have implemented the option for `Circle` to accept ellipses axes input, as well as the graphical representation of the ellipses.

But you have not implemented a magnetic field expression for the ellipses.

The library is based on analytical expressions (not discretized curves) so an analytical expression for the magnetic field computation of the `Ellipses` is required for it to yield a correct `getB()` output - similar to the core function `magpy.core.current_circle_Hfield()`.

Do you have such an expression?

PR #888, opened 2025-08-20
Force integration

OrtnerMichael:
Finally this seems to work. This is the last part of the NumFocus grant and will completely replace the `magpylib_force` package.

Here is how it works: Access force computation at top level:
```
F,T = magpy.getFT(sources, targets, pivot="centroid", eps=1e-5, meshreport=False)
```

- `sources` are any magnetic field generatorsc (single object or 1D list)
- `targets ` force and torque acting on targets is computed (single object or 1D list)
- `pivot` The Force adds to the Torque via the pivot point. For a freely floating magnet this would be the barycenter (= centroid when the density is homogeneous). If pivot="centroid" the centroid is selected as the pivot point for all targets. If pivot=None no pivot is used. This will give nonphysical results. If pivot=array_like of shape (3,) the same pivot is used for all targets. Alternatively one can provide an individual pivot point for each target.
- `eps` is the FD stepsize
- `meshreport` prints the target mesh size giving you an idea of the computation effort

Targets must have the `meshing` property defined. This is simply an int number which defines the target number of mesh points. The current meshing algorithms will not always create exactly the given target number, but aim to create uniform meshes with aspect-ratio = 1 cells.

@fslanovc, @Alexboiboi  and @luizenger please give it a try and give me feedback on the interface - it should be very intuitive to magpylib users. Here is a simple example:

```python
import numpy as np
import magpylib as magpy

cube = magpy.magnet.Cuboid(
    dimension=(1, 1, 1),
    polarization=(.1, .2, .3)
)
circ = magpy.current.Circle(
    diameter=2,
    current=1e3,
    position=(0, 0, 1),
    meshing=50,
)
F, T = magpy.getFT(cube, circ)

print(f'force: {np.round(F, decimals=2)} N')
# force: [ 13.65  27.31 -81.93] N

print(f'torque: {np.round(T, decimals=2)} Nm')
# torque: [-8.55  4.27 -0.  ] Nm
```

If any of you mange to give me a code review it would be appreciated.


OrtnerMichael:
Its working 💯  ... I'll add the docs (1-2 weeks) and then we are done.

OrtnerMichael:
the new Logo

image

fslanovc (approved):
approved! :)

PR #906, opened 2025-10-05
Add Current Orientation to TriangleStrip and TriangleSheet visualizations

Alexboiboi:
### TriangleStrip
image

### TriangleSheet
image


OrtnerMichael (approved):
Looks good to me. I fixed the discussed issue in the current_sheet example

Issue #883, opened 2025-08-07
Make getB more batchable


purepani:
Currently, it seems that `magpy.getB` is pretty restricted in how it can work with shapes. Based on the documentation, it can essentially only work with one extra dimension. It would be useful to be able to work with any number of extra dimensions. That way, you can easily calculate a ton of field values at once with broadcasting. 
For example, the following is currently not possible, but should be:

```python
import magpylib as magpy
import numpy as np

positions = np.arange((10, 20, 3))
B = magpy.getB(
     source='Cylinder',
     observers=positions,
     polarization=(0, 0, 1),
     dimension=(1, 1)
) # shape is (10, 20, 3)

positions = np.arange((10, 3))
polarization = np.arange((20, 3))
dimension = np.arange((30, 40, 2))
B = magpy.getB(
     source='Cylinder',
     observers=positions[:, None, None, None],
     polarization=polarization[None, :, None, None],
     dimension=dimension[None, None, :, :]
) # shape is (10, 20, 30, 40, 3)

```

I am guessing that the reason is due to the fact that the same limitation applies to `scipy.spacial.transform.Rotation` class, but this should be able to be worked around with reshaping before and after rotation.

OrtnerMichael:
HEy @purepani 

Thanks for pointing this out. We were too lazy - the original idea was that the users could do the reshaping - ofc. boadcasting i your example offers memory advantages. When moving to more array efficiency we should certainly consider this.

But also I am currently quite unhappy with the structure of `getB`. I think it the functional part should be separated from the oo one completely for obvious reasons (terrible docstrings, overloding confuses, ...). Maybe something using classmethods
```
magpy.magnet.Cuboid.getB(
    obs_array,
    pos_array,
    mag_array,
)
```

instead of 
`getB("Cuboid", ....)`

what do you think ? I'm also discussing this with myself in #704 :)

OrtnerMichael:
What you are suggesting is an improvement from a developer point of view - makes sense.

What do you mean by `getB_cuboid` ? You think the way its currently done is fine from a user perspective, or you think a function 'getB_cuboid()', `getB_sphere()` ... would be nice to have (top level?) functions for the functional interface ?

purepani:
There's also maybe a way to do this with typing generics but again, I feel like it's maybe trying to be too smart.

PR #916, opened 2025-10-29
🚀 Extend Path Logic to Object Attributes

Alexboiboi:
# 🚀 Extend Path Logic to Object Attributes

> **⚠️ IMPORTANT**: Due to the extensive changes across the codebase, this PR conflicts easily. It is currently **behind `main`** (#967 and #970 have since merged) and needs `main` merged in — 4 conflict hunks across 3 files, all mechanical.
>
> A **major version bump to v6.0.0** was originally assumed. It is **not needed**: `position` and `orientation` remain eagerly coupled as before, and padding is just-in-time rather than fully lazy, so the change is purely additive. See "Compatibility" below.

https://github.com/user-attachments/assets/64c8fa40-e215-446d-8958-a20081944c14

## Summary

This PR extends the existing path logic framework (currently available for `position` and `orientation`) to support dynamic paths for various object-specific attributes. This enables path-dependent or parametric variations of physical properties beyond just motion and rotation, unlocking powerful new use cases for magnetic field simulation and visualization.

## Motivation

The current path system provides powerful functionality for animating object positions and orientations. Many real-world scenarios require dynamic changes to other physical properties:
- Path-varying currents in coils
- Deforming coil geometries (e.g., moving vertices in Polylines)
- Changing coil dimensions (e.g., expanding/contracting Circle diameter)
- Path-dependent magnetization and magnet dimensions

## Key Technical Concepts

The path model is **lazy storage + just-in-time (JIT) padding** (not pure-lazy broadcasting — see Future Work):

**Lazy storage**: Each property is kept in its minimal user-given form (length 1 unless explicitly path-varying). `move()`/`rotate()` only touch `position`/`orientation` and leave other path properties minimal, avoiding cascading side-effects when properties are set independently.

**Just-in-time padding**: At field/force/display time the minimal arrays are edge-padded up to the common max path length — *materialized copies, not broadcast views* — then restored afterwards (via `_preserve_paths`), keeping the synchronization transparent to the user.

**Eager sync for position/orientation only**: `position` and `orientation` remain eagerly synchronized with each other (as before), since geometric consistency is always required.

**`BaseGeo._sync_path_lengths(prop_names)`**: JIT helper that edge-pads each named property to the common max path length and returns a dict of freshly-materialized arrays, ready for injection into field/force computation. The object's own minimally-stored attributes are left untouched.

**Extensible design**: Each class declares supported path properties via the `_path_properties` tuple, automatically aggregated from parent classes.

### ✅ Completed

- [x] **Core infrastructure**
  - [x] Implement `_path_properties` with automatic aggregation via `__init_subclass__`
  - [x] Implement just-in-time `_sync_path_lengths()` on `BaseGeo`
  - [x] Implement `pad_path_properties()` utility
  - [x] Eager sync between `position` and `orientation` (geometric consistency)

- [x] **Input validation framework**
  - [x] `match_shape()` with ellipsis support
  - [x] `check_condition()` for value validation
  - [x] `check_format_input_numeric()` comprehensive validator

- [x] **Current Sources** — Polyline, Circle, TriangleStrip, TriangleSheet
- [x] **Magnet Sources** — Cuboid, Cylinder, CylinderSegment, Sphere, Tetrahedron, TriangularMesh
- [x] **Misc Sources** — Dipole, Triangle

- [x] **Field computation** — path properties tiling in `_getBH_level2()`
- [x] **Force/Torque computation** — path-aware mesh generation for all source types

- [x] **Testing**
  - [x] `test_force_path_varying.py` — comprehensive path-varying force tests (4 sections: current sources, magnet sources, collections, misc)
  - [x] Mismatched path-length regression tests for all target classes
  - [x] Regression tests for orientation setter, dipole moment, vertices setter, and degenerate triangle handling

- [ ] **Misc Sources** — CustomSource (field function parameters)
- [ ] **Documentation** — user guide, examples, migration guide
- [ ] **Performance** — profile and optimize path sync overhead
- [ ] **Pure-lazy broadcasting** (future) — replace JIT padding copies with broadcast views passed directly into the field kernels, eliminating the pad/restore step; best coordinated with the Array API work (#883)

## Use Cases Enabled

1. **Path-varying currents**: Simulate AC coils, current ramps, or switching sequences
2. **Deforming geometries**: Model coils that change shape during motion
3. **Dynamic sizing**: Simulate thermal expansion, mechanical deformation
4. **Complex animations**: Visualize objects changing multiple properties simultaneously
5. **Isolated property studies**: Compute fields with only property changes (no motion)

**Example - Expanding coil with increasing current:**
```python
import magpylib as magpy
import numpy as np

diameters = np.linspace(0.01, 0.05, 10)  # 1cm to 5cm
currents = np.linspace(1, 10, 10)         # 1A to 10A
circle = magpy.current.Circle(diameter=diameters, current=currents)
circle.move([(0,0,z) for z in np.linspace(0, 0.1, 10)])

B = magpy.getB(circle, observers)
```

**Example - Complex deforming spiral with varying current:**
```python
import magpylib as magpy
import numpy as np

pos_len = 20
vert_len = 50
curr_len = 100
vert_num = 100
omega_curr = 10
omega_vert = 2
lsc = np.linspace(0, np.pi, curr_len)
lsv = np.linspace(0, np.pi, vert_len)
verts = []
for zz in np.sin(omega_vert * lsv) / 2:
    r = np.linspace(1, 0.2, vert_num)
    t = np.linspace(0, 4 * 2 * np.pi, vert_num)
    x = r * np.cos(t)
    y = r * np.sin(t)
    z = x * y * zz
    verts.append(np.array([x, y, z]).T)
verts = np.array(verts)
curr = np.sin(omega_curr * lsc) * 20
pos = np.linspace([0, 0, -0.5], [0, 0, 0.5], pos_len)
line = magpy.current.Polyline(position=pos, vertices=verts, current=curr)
sens = magpy.Sensor()
with magpy.show_context(
    line,
    sens,
    animation=True,
    style_arrow_size=4,
    style_arrow_color="red",
    animation_time=5,
    animation_maxfps=30,
) as ctx:
    ctx.show(output="model3d", col=1)
    ctx.show(output="Bz", col=2)
```

## Also in this PR (beyond the path system)

- **`describe()` / `_repr_html_` / `__repr__` moved onto `BaseGeo`.** `class_BaseDisplayRepr.py`, `class_BasePropDipole.py` and `class_BasePropVolume.py` are deleted (−184 lines) and folded into the base class. No behavioural change intended.
- **New public `obj.path_properties`** — lists which of an object's attributes support a path dimension.
- **Numeric input validation rewritten** around a single `check_format_input_numeric` with shape-pattern matching (wildcards, ellipsis) and value conditions, replacing the per-attribute validators.
- **`magpylib._src.utility.check_path_format` removed** — path consistency is now maintained by the path-property machinery itself.

## Compatibility

Verified against `main`:

- Return shapes for non-path-varying objects are **unchanged** (`dipole_moment` → `(3,)`, `volume` → scalar, `centroid` → `(3,)`).
- Derived properties gain a leading path axis **only** when the inputs they derive from are path-varying (`dipole_moment` → `(p, 3)`).
- Objects whose properties have differing path lengths are edge-padded to the longest, rather than raising.

**No breaking changes.** `position` and `orientation` stay eagerly synchronized with each other exactly as before, and path padding is just-in-time rather than fully lazy, so existing code sees the same behaviour it always did. The changelog entry therefore has no "breaking changes" section, and this can ship as a minor release.

Pure-lazy broadcasting — replacing the JIT padding copies with broadcast views passed straight into the field kernels — is future work and *would* be the point at which compatibility needs revisiting.


purepani:
I think a PR like this would be easier to make if we were to make each level broadcastable in all parameters(https://github.com/magpylib/magpylib/issues/883). That way, it would maybe avoid some extra memory usage needed with tiling stuff constantly, and the code could probably be reduced. That's something that could be done either along with the array api stuff or independently from it.

Alexboiboi:
👋 @purepani 
Thanks for chiming in!
Could you be a bit more specific on broadcasting, and how it could fit here?

Alexboiboi:
> If the core functions looked like
> 
> ```
> # (..., 3), (...,3) -> (..., 3)
> def dipole(x: np.array, m: np.array): ...
> ```
> 
> then all the properties of the classes can be batched as well:
> 
> ```
> class Dipole:
>      x: np.array # (..., 3)
>      m: np.array # (..., 3)
> ```
> 
> You can then pass a `time_axis` argument to the initializer to specify an axis to use as a time axis, and then have special semantics based on that. In a vacuum, that's how I would imagine doing this.

Thanks for the feedback @purepani!

I appreciate the broadcasting approach you're suggesting - it's definitely more aligned with NumPy conventions and could reduce memory overhead from padding/tiling. However, implementing it would require:

- Rewriting all field computation functions to handle arbitrary batch dimensions
- Redesigning the core path synchronization logic
- Updating how observers/sources interact with batched properties
That's a much larger architectural change than the current PR scope. For now, I'd like to keep the explicit padding approach because:

- It's simpler to reason about (all path properties guaranteed same length)
- It maintains consistency with the existing position/orientation path behavior
- It works well for the common use case where path lengths are similar anyway
I think the broadcasting approach would be an excellent candidate for a future refactoring PR, possibly coordinated with the Array API work you mentioned in #883. We could design it properly from the ground up rather than trying to retrofit it here.

Does that sound reasonable? Happy to discuss further!

purepani:
I think that makes sense. I was thinking that it would be a good idea to think about if 
A. If the suggestion I made should be a long term goal
B. If yes, then if there's anything to consider for the design here. For example, maybe the suggested breaking change isn't necessary since broadcasting rules would render the same as the current situation(at least if I'm thinking about this correctly).

purepani:
I wouldn't necessarily remove the functional interface. It's a much more "direct" way of doing the computations rather than through the object(and it at least personally meshes with the way I think better, so this is a bit of a selfish request).

As for a practical reason to keep it, if you want to use the array api, and in particular, jax, the easiest way to get jit compilation to work well is through the functional interface. Getting it to work with objects is possible, but requires making the classes pytrees; that's been a bit of a problem for the scipy Rotation class array api implementation because you need to import jax to make them pytrees, so for that reason, I would recommend against getting rid of the functional interface for now.

OrtnerMichael:
Also - if we are doing this, we should not refer to it as path computation, but as batch computation, and rework the docs distinguishing

**single computaion** VS **batch computation**

Unless you have better naming ideas - but batch feels ok, and has no ambiguity - and could be extended to parallel processing.

purepani:
Let me see if I can pull out some of the code from my PRs and make the core functions batch-friendly. I remember doing some of the work but not all of it. It's not breaking and it should be pretty easy to do.

Alexboiboi:
@OrtnerMichael 
as agreed in last private discussion, I am merging this now. Some changes can be made before a release anyway

PR #920, opened 2025-11-23
Fix spelling in comments in test_elliptics.py

MischaMegens2:
'vectroized implemetnation' seemed like a typo, this is a fix.


OrtnerMichael (approved):
thanks for the fix !!!

Issue #922, opened 2025-11-26
(In)consistent vacuum magnetic permeability mu_0


MischaMegens2:
Since CODATA19, the vacuum permeability $\mu_0$ is no longer exactly equal to $4 \pi 10^{-7}$ H/m. This has been addressed in Magpylib by referring to Scientific Python's MU0, see Issue #731 [the new MU0 problem](https://github.com/magpylib/magpylib/issues/731). However, it seems there are still some $4 \pi 10^{-7}$ lingering, for instance in:
`line #91 in field_BH_circle.py:`
`    return np.vstack((Hr, np.zeros(n5), Hz)) * 795774.7154594767  # *1e7/4/np.pi`
`line #364 and #389 in class_BaseExcitations.py:`
`    self._polarization = self._magnetization * (4 * np.pi * 1e-7)`
`    self._magnetization = self._polarization / (4 * np.pi * 1e-7)`
Some of the tests have it too.
It would be nice to change these, and use consistent value for the permeability.

MischaMegens2:
I'd propose:
`pf = i0 / (4 * np.pi * r0 * np.sqrt(x0) * q2)`
`return np.vstack((Hr, np.zeros(n5), Hz))`
The prefactor `pf` matches the expression in [Michael Ortner _et al_., Numerically Stable and Computationally Efficient Expression for the Magnetic Field of a Current Loop](https://www.mdpi.com/2673-8724/3/1/2), Eq. 17a:
$B_z = B_0 {k \over q^2 \sqrt{\bar{\rho}}} cel^{\*\*}(\bar{\rho},k,q)$
Substituting $B_0 = {\mu_0 i_0 \over 8\pi\rho_0}$; $B=\mu_0 H$; and $k^2 = {4\bar{\rho} \over \bar{z}^2 + (1+\bar{\rho})^2} \equiv {4 \bar{\rho} \over x_0}$, this becomes
$H_z = {i_0 \over 8\pi\rho_0} {2 \over q^2 \sqrt{x_0}} cel^{**}(\bar{\rho},k,q)$
This confirms that the proposed prefactor is indeed correct (provided that we ignore the missing sign, which is not our concern here).

OrtnerMichael:
Hey @MischaMegens2,

Yes, you are correct about this one - thanks for catching yet another bug - or code smell in this case !
Since we are canceling the MU0s the tests should also not fail.

I have to confess that many implementations have grown over time so there are probably more issues like this one left in the code.

[...]

PR #923, opened 2025-11-26
Use a consistent value for vacuum magnetic permeability mu_0

MischaMegens2:
Replace hard-coded mu_0 values by reference to mu_0; addresses Issue #922


Alexboiboi:
Some `describe` tests are failing because of slightly different property value changes. The `describe` logic has been reworked in #916, so not point to fix the failing tests here. I'll merge as is.

Thanks ;)

PR #931, opened 2025-12-26
Fix current_circle_Hfield() dependence on mu0

MischaMegens2:
Issue #922: Fix prefactor `pf`: remove `/ 20 * 1e-6` (i.e., $\mu_0/8\pi$), cancel it against `/ MU0`, so `current_circle_Hfield()` becomes independent of $\mu_0$ altogether.

### DIFF
--- a/tests/test_vs_mag2.py
+++ b/tests/test_vs_mag2.py
@@ -35,8 +35,13 @@ import magpylib as magpy
 
 def test_vs_mag2_linear():
     """test against magpylib v2"""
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):

OrtnerMichael:
now the rep will see a minor difference in representation on the last commas ... i can fix that one

Alexboiboi:
> i can fix that one

thanks but like already mentioned, no need to fix it now, this is reworked anyway in #916 ;)

Issue #932, opened 2026-01-03
Divide by zero in current_circle_Hfield() on axis


MischaMegens2:
The function `current_circle_Hfield()` runs into a series of 'divide by zero' warnings when trying to evaluate the magnetic field on axis (of a current loop); and as a consequence, it does not provide the right result for the H field, e.g.,
```
import numpy as np
import magpylib as magpy

H = magpy.core.current_circle_Hfield(
    r0=np.array([1]), 
    r=np.array([0]),
    z=np.array([0]), 
    i0=np.array([1])
)
print(H)
```
results in
```
[[nan]
 [ 0.]
 [nan]]
[magpylib\src\magpylib\_src\fields\field_BH_circle.py:82] (file:///C:/magpylib/magpylib/src/magpylib/_src/fields/field_BH_circle.py:82):
RuntimeWarning: divide by zero encountered in divide
  Hr = pf * z / r * _cel_iter(q, p, np.ones(n5), cc, ss, p, q)
```

This does not pose a problem for magpylib in practice, since `_BHJM_circle()` catches `r == 0` (on-axis) as a special case and evaluates the field without relying on `current_circle_Hfield()`. However, the division by zero there can be easily remedied, and the aesthetically not so pleasing special case avoided altogether. The remedy is to notice that the division by $r$ can be canceled against a factor with $r$ in the numerator; for instance, in the prefactor $k \over \sqrt{\bar\rho}$ of the $cel()$ functions, $k^2 = {4 \bar\rho \over x_0}$, so we can rewrite the prefactor as $2 \over \sqrt{x_0}$ (this has been fixed already in #931). Slightly more challenging is the prefactor $z \over r$ in $H_r$; it can be remedied by noticing that $cel(c,s)$ as a function of $c$ and $s$ is homogeneous: $cel(\lambda c, \lambda s) = \lambda ~ cel(c,s)$, so the $z \over r$ can be moved inside the function arguments, and the $r$ (or $\bar \rho$) then cancels against the numerators $k^2$ there. It does not affect the convergence in any way, apart from avoiding the 'divide by zero' warnings.
The `current_circle_Hfield()` then gives the right answer on-axis too. It seems it would be a beneficial change.

PR #933, opened 2026-01-04
Address Issue #932

MischaMegens2:
- Fix divide by zero in current_circle_Hfield(), so it works on-axis too
- Remove special case in  _BHJM_circle(), since it is no longer needed
- Add tests

### DIFF
--- a/src/magpylib/_src/fields/field_BH_circle.py
+++ b/src/magpylib/_src/fields/field_BH_circle.py
@@ -24,7 +24,7 @@ def current_circle_Hfield(
     Parameters
     ----------
     r0 : array-like, shape (i,)
-        Loop radii.
+        Loop radii, should be positive (r0 > 0).
     r : array-like, shape (i,)
         Radial observer positions.
     z : array-like, shape (i,)
@@ -60,7 +60,6 @@ def current_circle_Hfield(
      [0.077 0.226]]
 
     """
-    assert np.all(r0 > 0), "Zero loop radius is not allowed"
     n5 = len(r)
 
     # express through ratios (make dimensionless, avoid large/small input values, stupid)


MischaMegens2:
> Hi @MischaMegens2,
> 
> So this looks like a good solution that improves the flow and readability by a lot!

Glad that you liked it
 
> But the `assert` in Line 63 in field_BH_circle is not a good idea with `np.all()` being a very "slow" operation. 

Oh, I wouldn't consider it 'very slow', I'd expect the next few lines to take at least as much effort, if not more:
```
r = r / r0
z = z / r0
```
Moreover, the 'assert' is ignored in Optimize mode (if you need the speed). (My reason to put in the assert is more to confirm the obvious expectation, [Design-by-Contract](https://en.wikipedia.org/wiki/Design_by_contract)-style.)
But then again, I am not very attached to the `assert` either, and those two next lines will throw a warning in any case. Perhaps the 'let it fail' approach is more appropriate for a Python library.

> We thought that core users would be quite advanced and would have to know what they are doing - while we focus on making the object oriented interface (99% of users) as good as possible. One could think to wrap the core function into another interface function that catches such input-faults [...]

I wouldn't suggest it. We don't need yet another level of interface function. Clearly it is overkill.

> This should certainly be remaked up in the function docstring.
> What is your opinion?

I think that's a good suggestion. Let me try and put in something.

OrtnerMichael:
Hey @MischaMegens2 

You are right - I had it in my head from benchmarking that `np.all()` is really slow, but apparently its not :). This was my greatest concern.

If it takes practically no time compared to the rest I'd be happy to keep the `assert` - silent errors are the worst for a developer. There would be a lot of asserts to add all over the core functions, to make it consistent. I'll start an issue on this.

PR #941, opened 2026-02-15
Refactor complete elliptic integrals routines _cel

MischaMegens2:
- Adds test_cel_Bulirsch: checks _cel() results against values published in R. Bulirsch, Numerische Mathematik 13, 305-315 (1969), section 4.2.
- Refactors _cel0 step-by-step, in a series of commits that make it shorter, more python style, and with notation closer to the Bulirsch paper. In particular, it avoids repeating the main algorithm twice, by un-unrolling the iteration loop, as it were.
- Likewise refactors the vectorized version _celv (in one commit), and makes it conform to _cel0.

It passes the tests as before.

### DIFF
--- a/src/magpylib/_src/fields/special_cel.py
+++ b/src/magpylib/_src/fields/special_cel.py
@@ -13,36 +13,36 @@ def _cel0(kc, p, c, s):
     R. Bulirsch, Numerical Calculation of Elliptic Integrals and Elliptic Functions. III
     Numerische Mathematik 13, 305-315 (1969).
     """
-    if kc == 0.0:
+    if kc == 0:
         msg = "FAIL cel: kc==0 not allowed."
         raise RuntimeError(msg)
     errtol = 0.000001
-    if p > 0.0:
+    if p > 0:
         p = np.sqrt(p)
         s = s / p
     else:
         f = kc * kc
-        q = 1.0 - f
-        g = 1.0 - p
+        q = 1 - f
+        g = 1 - p
         f -= p
         q *= s - c * p
         p = np.sqrt(f / g)
         c = (c - s) / g
         s = -q / (g * g * p) + c * p
-    mu = 1.0
+    mu = 1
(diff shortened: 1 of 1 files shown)


MischaMegens2:
> So this looks really good - just one question, have you benchmarked this implementation - in the sense of speed gain/loss and maybe memory ? If I remember correctly I implemented the version from Derby et.al ... not sure if it makes a difference.

As for benchmarking: I have tried timing it, like
```
import magpylib as magpy
import timeit
display(timeit.timeit('magpy._src.fields.special_cel._cel0(0.5, 1., 1., 1.)', globals=globals()))
```
This does not show a significant difference between the original and refactored implementations.

I would not expect much difference, since the essence of the algorithm is unchanged, and the speed/memory benefit of python things like augmented assignment (e.g., '+=') is minor (avoiding an occasional reallocation of memory is not going to make a large difference). My main intention with the refactoring is more to simplify the implementation (avoiding 'unrolling the loop' to not repeat oneself, making it shorter; using fewer variables, with naming closer to that in the Bulirsch paper) in an attempt to make it easier to understand what is going on.

As for where the original implementation came from: the cel() docstring mentioned 'vom Kirby2009', but I'm not quite sure what that refers to. The cel() routine in your paper 'Numerically Stable and Computationally Efficient Expression
for the Magnetic Field of a Current Loop' mentions 'Bulirsch’s cel algorithm taken from Derby2010', and the cel() in your paper is otherwise substantially the same as the 'vom Kirby' cel() in magpylib. It does match the BASIC code in Derby2010, which has unfortunate naming like `k = ABS(kc)` (whereas $k^2=1−k_c^2$ is something else elsewhere in the Derby paper, and in other literature). 
So I would suggest staying closer to the Bulirsch1969 version, short and sweet.
The incomplete elliptic integral functions in 'special_el3.py' also follow Bulirsch' implementation directly, so that would be consistent.

Issue #935, opened 2026-02-09
minor code improvement


OrtnerMichael:
this is from field_BH_current_sheet ... can this not be done in a single "xzx" rotation ?

```
# step 2
    # apply two rotations so that Pi2 -> (u1, 0, 0)

    # step 2.1: first rotation around x-axis so that Pi2 -> xy-plane
    theta = -np.arctan2(vertices[:, 1, 2], vertices[:, 1, 1])

    r21 = R.from_euler("x", theta[:, np.newaxis])

    vertices[:, 1, :] = r21.apply(vertices[:, 1, :])
    vertices[:, 2, :] = r21.apply(vertices[:, 2, :])

    # step 2.2: second rotation around z-axis so that Pi2 -> x-axis
    alpha = -np.arctan2(vertices[:, 1, 1], vertices[:, 1, 0])

    r22 = R.from_euler("z", alpha[:, np.newaxis])

    vertices[:, 1, :] = r22.apply(vertices[:, 1, :])
    vertices[:, 2, :] = r22.apply(vertices[:, 2, :])

    # step 3
    # apply rotation around x-axis so that Pi3 -> (u2, v2, 0)
    psi = -np.arctan2(vertices[:, 2, 2], vertices[:, 2, 1])

    r3 = R.from_euler("x", psi[:, np.newaxis])

    # can this not be done in one "xzx" rotation?
```

PR #950, opened 2026-04-14
Combine three rotations into single xzx Euler in current sheet transform

stanbot8:
Addresses #935, where @OrtnerMichael asked whether the three sequential rotations in `_coordinate_transformation` could be combined into a single "xzx" rotation.

The three sequential rotations (x, z, x) are now built as a single `R.from_euler('xzx')` rotation. The second angle is computed with direct cos/sin on the untransformed vertex instead of applying the first rotation, reducing the number of `Rotation.apply()` calls from 5 to 2.

**Before:**
- 3 `R.from_euler` constructions, 5 `.apply()` calls
- Vertices mutated in-place at each step

**After:**
- 2 `R.from_euler` constructions (`'xz'` for angle calc, `'xzx'` combined)
- 2 `.apply()` calls (one per vertex, at the end)
- Second angle computed with `cos(theta)`, `sin(theta)` directly

**Verification:**
- 1072 non-display tests pass unchanged (including `test_core_physics_current_sheet_VS_Polyline` which cross-checks field output against the independent Polyline implementation)
- Numerical equivalence to the original across 1000 random triangles: coordinates match to 5e-15, rotations match to 2e-15 (machine epsilon)
- Ruff clean, Pylint 10.00/10

### DIFF
--- a/src/magpylib/_src/fields/field_BH_current_sheet.py
+++ b/src/magpylib/_src/fields/field_BH_current_sheet.py
@@ -44,36 +44,39 @@ def _coordinate_transformation(vertices):
     vertices[:, 2, :] = vertices[:, 2, :] - translation
     vertices[:, 0, :] = 0
 
-    # step 2
-    # apply two rotations so that Pi2 -> (u1, 0, 0)
+    # Three sequential rotations (x, z, x) place P2 on the x-axis and
+    # P3 in the xy-plane. Each angle depends on the previous rotation's
+    # result, but we compute them with direct trig (no intermediate
+    # .apply() calls) and apply the combined rotation once at the end.
 
-    # step 2.1: first rotation around x-axis so that Pi2 -> xy-plane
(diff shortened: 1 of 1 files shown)


OrtnerMichael:
Hey @stanbot8,

Thanks a lot for picking up this issue and sorry for the slow reply.
I'll review promptly - did you also make performance tests ?

br michael

OrtnerMichael:
Hey @stanbot8,

So I went through your code, and tested it thoroughly.

The results look good, but there is no real performance gain.

I asked codex to come up with a solution - and it suggests relying on matrix multiplications instead of rotation objects. It produced some code that it long and gives correct results and is three times faster than the original and your implementation (at least when array size exceeds 100 - the bottleneck beiing the creating of the rotation object.

Without significant benefit we will not merge this change. Please let me know if you think I overlooked something.

PR #951, opened 2026-04-17
Use cel() instead of cel_iter()

MischaMegens2:
The calculation of the magnetic field near a circular current loop involves the evaluation of combinations of complete elliptic integrals; a straightforward implementation suffers from numerical cancellation [[1]]. To avoid this, Magpylib uses a combination of two strategies: Bulirsch's cel integral (instead of subtracting Legendre's integrals), preceded by an analytical Bartky/Landen transformation. To implement this, the original `cel` algorithm was split up, into a `celx/xx` part (for various components of the field) and a `cel_iter` part (for the remaining iterative part of `cel`).

However, we don't actually need a separate `_cel_iter()`, the same effect can be achieved using `_cel()` directly, by scaling its arguments and its result a bit. This shortens `special_cel.py` considerably.

The graphs below are intended to demonstrate that the proposed approach performs equally well; these show the relative accuracy of the radial ($B_r$) and axial ($B_z$) components of the field, obtained by various methods (smaller/more blue is better):

image
image
image
image
image

[For definiteness, the python scripts to generate the plots above are attached [here](https://github.com/user-attachments/files/27141722/scripts.for.plots.in.pull.request.txt).]

Clearly 'cel+Landen' works equally well as 'cel_iter'.
It brings the advantage of a much shorter `special_cel.py`.

[1]: https://doi.org/10.3390/magnetism3010002


OrtnerMichael:
hi @MischaMegens2,
thanks for this - I'll be on it soon - sorry for the wait!

OrtnerMichael:
hey @MischaMegens2,

I finally managed to review your suggestion. Thanks a lot for your work.

**So here is how I understand it:**

You reformulate `current_circle_Hfield()` to make use of the already existing `cel()` algorithm, which enables you to dump the extra `cel_iter()` algorithm. This reduces code and bases current loop computation and cylinder computation on the same algorithm. Good motivation :).

Here is the catch - When I developed the `current _circle` part, I noticed that you can save a lot of computation time by formulating it like I did. If you compare the computation times, I think you will find that it is in general about 3 times faster for array sizes 1e1 to 1e6.

However - this is the result of the `cel` algorithm making use of an iteratively shrinking mask, which sounds like a great idea but seems to require more effort in evaluating and reducing the mask than it saves from the reduced computation. I guess this is where numpy is truly limited.

After fixing the `cel` algorithm your suggested computation is about 20-40 % slower than mine for array sizes until 1e3 and about 10% for larger arrays.

**Conclusion:**
Did you check computation performance ? Do you agree with this assessment ?
The cel algorithm must be fixed - I'll do that in the next days. I'd prefer to unify ... but maybe the unification should be around `cel_iter` and not `cel`. For this I have to look in more detail into the Cylinder codes...

...... I had a look at those codes now. cel is called from cel3 and from the cylinder functions. I don't think they will benefit from calling to cel_iter directly. this would also reduce readability as the comparison with literature becomes difficult.

OrtnerMichael:
So i fixed the cel functions in this PR: #955.

I tried to merge `cel_iter` with `cel` but that does not work very well, as `cel_iter` is designed to step in after the first iteration which adds another few lines to `cel` and generally makes the `cel` implementation much less readable / comparable to the literature.

GPT suggests to have them as separated functions as they are now.

br michael

MischaMegens2:
Hi @OrtnerMichael,

Ah that's interesting!
I had just presumed that the more elaborate implementation of `cel()` would be the faster one, relative to `cel_iter()`. But it doesn't require a stretch of the imagination to believe that the messing around with the `[msk]` subscripts doesn't make it any faster.
Let me think about it for a bit.

Best,
Mischa

Alexboiboi:
Hi @MischaMegens2 and @OrtnerMichael 

what is the status on this PR ? Do we want to add this work to the repo at some point#? If so this branch needs to get up to date with main because  `special_cel.py`  moved under it accrosss #941, #945, #955 and #958. 
@MischaMegens2 I think this is best if you do this yourself, as you know where this is going, thanks ;)

MischaMegens2:
Hi @Alexboiboi 

I would like to come back to #951, but I think #959 should go first (could you take a look at that?), and I understood from @OrtnerMichael  that some bigger changes were coming over the summer, so I was waiting for that (and focusing on some other things in the mean time... but it is on my list still)

Alexboiboi:
@MischaMegens2 
Yes, these two prs should be merged before the array-api work continues I think

MischaMegens2:
> @MischaMegens2 ...these two prs should be merged before the array-api work continues I think

The #959 is good to go in my opinion
But the #951 will need some more TLC, if it's suitable at all (@OrtnerMichael observed that it is slower than the original, so it might not be such a good idea as it stands...)

PR #958, opened 2026-05-21
Improve cel_iter_*

MischaMegens2:
A closer look at `cel_iter_...()`, after the revision in #955, suggests two minor improvements/simplifications:

- The definitions for `cel_iter_scalar()` and `cel_iter_vector()` appear almost identical - the only differences are the `m` vs. `np` libraries, and a call to `np.any()`. It does not take much to parametrize this, so we don't have to repeat ourselves.
- In `cel_iter()`: replacing the for loop (with its pre-allocation and indexing) by a list comprehension is faster; it is how it was done in `cel()` before; and it avoids having to commit to `dtype=float`.

Here is a comparison of timings before and after these changes, for both scalar (`_CEL_SCALAR_THRESHOLD` large) and vector (`_CEL_SCALAR_THRESHOLD = 0`) style `magpy.core.current_circle_Hfield()`:

image

The plot shows that merging the scalar and vector function did not adversely affect the timings, so it is a nice cleanup.
(It also suggests slightly increasing `_CEL_SCALAR_THRESHOLD = 15`, as it was in `_cel()` before...)

For definiteness, the script that generates the plot is attached here: [cel_iter_improvements.ipynb](https://github.com/user-attachments/files/28079487/cel_iter_improvements.ipynb)

### DIFF
--- a/src/magpylib/_src/fields/special_cel.py
+++ b/src/magpylib/_src/fields/special_cel.py
@@ -1,7 +1,5 @@
 """Special functions cel."""
 
-# pylint: disable=too-many-positional-arguments
-
 import math as m
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


OrtnerMichael (approved):
This looks like a nice code cleanup ! Thank you !

PR #959, opened 2026-05-24
_cel() suggestions

MischaMegens2:
A few minor suggestions:

- In `_cel_scalar()`: avoid in-place update `s /= p`, keep `s = s / p` instead, since we don't want to modify `s` in caller: 
An in-place update could in principle modify a variable in a function's caller, as in:
  ```python
  def f(x):
      x /= 2
      return x
  
  x = np.array(1)
  print(f(x))
  print(x)
  ``` 
  (for instance, when `x` is a numpy array). 
  That's why in `_cel_vector()`, the in-place update `ss /= p` is preceded by `ss = s.copy()`.

  Admittedly this is not immediately an issue, since `np.isscalar(s) -> True`, `_cel_scalar` is intended for scalars.
  But to me, this is a little too close for comfort.
  And it turns out it is for no benefit: `s=s/p` is actually slightly faster than `s/=p` in the situation at hand:

  ```python
  import numpy as np
  import timeit

  display(timeit.repeat('s /= p', setup='s = np.float64(0.1); p = np.float64(1)', number=100000000, globals=globals()))
  display(timeit.repeat('s = s / p', setup='s = np.float64(0.1); p = np.float64(1)', number=100000000, globals=globals()))
  [3.8989724999992177,
   3.931931999977678,
   3.937487700022757,
   4.007137299980968,
   4.009531599935144]
  [3.860392199945636,
   3.8086202000267804,
   3.8343616999918595,
   3.8326475999783725,
   3.8462840999709442]
  ```
- `_cel_vector()`: `pp -> p, cc -> c, ss -> s`
We don't need reference to the original function arguments after copying,
we can shadow the original arguments `p, c, s` on purpose: `c = c.copy()`, etc.
and we stay closer to the original Bulirsch formulation and to `_cel_scalar()`, by using just `p, c, s` instead of introducing additional `pp, cc, ss`.
- Change the alignment of the formula lines in the `_cel()` docstring, so the parts of the integral sign line up
- `_cel()`: Use a list comprehension instead of a for loop with indexing: faster, and it avoids committing to `dtype=float` in the initialization of result 
(and it was like that before)
- `_cel_scalar()` stated: 'Caller ensures that kc != 0' (i.e., `_cel()` would take care of it):
  I'd suggest to include this in `_cel_scalar()`, since 
  1) `_cel_vector()` deals with the `kc==0` case internally too, it is consistent; 
  2) in `_cel()` it clutters up the list comprehension, while in `_cel_scalar()` it fits naturally at the beginning;
  3) it simplifies the `_cel_scalar` docstring (no 'Caller ensures that...')

### DIFF
--- a/src/magpylib/_src/fields/special_cel.py
+++ b/src/magpylib/_src/fields/special_cel.py
@@ -135,6 +135,7 @@ def _cel(kc: np.ndarray, p: np.ndarray, c: np.ndarray, s: np.ndarray) -> np.ndar
 
 # pylint: disable=too-many-positional-arguments
 
+
 def _cel_iter_scalarvector(qc, p, g, cc, ss, em, kk, xp=m, any_=lambda b: b):
     """
     Iterative part of the Bulirsch cel algorithm.
--- a/src/magpylib/_src/fields/special_cel.py
+++ b/src/magpylib/_src/fields/special_cel.py
@@ -10,12 +10,14 @@ _CEL_ERRORTOL = 1e-8
 
 def _cel_scalar(kc, p, c, s):
     """
-    scalar version of bulirsch cel algorithm
-    caller ensures that kc != 0
+    Scalar version of Bulirsch cel algorithm
     """
+    if kc == 0:
+        return np.nan
+
     if p > 0:
         p = m.sqrt(p)
-        s /= p
+        s = s / p
     else:
         f = kc * kc
         q = 1 - f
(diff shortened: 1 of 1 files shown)


MischaMegens2:
@OrtnerMichael:
Would you have a chance to take a look at the Pull Request?  This one should be easy (I hope).

Alexboiboi:
Hi @MischaMegens2

This `_cel` edits looks good to me and ready to merge but I'd still like @OrtnerMichael 's stance on it before we squash-merge this.

I rebased this onto main and force-pushed, so you'll want to reset your local copy (if need be).

Btw, also timed the list comprehension change: kc.size=1 is about 0.25us slower per call on my machine (irrelevant in my opinion), kc.size=9 slightly faster. The vectorised branch is renames only so nothing to measure there.

MischaMegens2:
Hi @Alexboiboi 

Thanks for the heads-up; I'll just wait for @OrtnerMichael to take a look then,

Best,
- Mischa

PR #985, opened 2026-08-26
Make the docs wordmark magnetic, and give it a magpie

Alexboiboi:
Docs-only. The wordmark beside the logo becomes magnetic — each letter a dipole, the pointer a bar magnet, using the same physics the library computes — and the magpie in the logo comes to life to fetch letters back. Adds a 404 page.

## Demo

https://github.com/user-attachments/assets/47202cea-acfc-46a5-b9c9-1f2f2fc9add9

## Scope

Everything is under `docs/`, plus one line in `pyproject.toml`: `sphinx-notfound-page` in the `docs` extra, needed by the 404 page.

image

## Opting out

A recording won't show these, so:

- `prefers-reduced-motion: reduce` — nothing is built at all.
- Touch or no hover — the wordmark keeps its colours, nothing moves, and the logo keeps its painted magpie.
- Click the magpie and it goes back into the logo for good; click the same spot and it comes back out (`localStorage`).
- It helps itself to a letter once a visit, only once the reader has genuinely stopped, and stashes it bottom-right rather than on the text.
- The bird is `pointer-events: none` and never swallows a click meant for the page under it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Alexboiboi:
I'll merge this now, if we don't like it we can always remove it ^^

